PR #23, opened 2022-05-27
fuzzing: fix tests

giuseppe:
adjust fuzzing tests to the new API.

### DIFF
--- a/fuzzing/main.c
+++ b/fuzzing/main.c
@@ -26,97 +26,46 @@ int LLVMFuzzerInitialize(int *argc, char ***argv)
 bool iter_cb(void *private, const char *name, int namelen, u64 ino, unsigned int dtype)
 {
 	struct test_context_s *test_ctx = private;
-	struct lcfs_context_s *ctx = test_ctx->ctx;
-	struct lcfs_inode_s cfs_ino_buf;
-	struct lcfs_inode_s *cfs_ino;
-	struct lcfs_inode_data_s cfs_ino_data_buf;
-	struct lcfs_inode_data_s *cfs_ino_data;
-	struct lcfs_dentry_s dentry_buf;
-	ssize_t i, size_xattrs;
-	char xattrs[256+1];
-	const char *cstr;
+	struct lcfs_xattr_header_s *xattrs;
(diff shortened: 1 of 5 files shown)


giuseppe:
@alexlarsson PTAL

fuzzing/main.c
```diff
@@ -26,97 +26,44 @@ int LLVMFuzzerInitialize(int *argc, char ***argv)
...
-	if (IS_ERR(cfs_ino))
-		return true;
-
-
-	cfs_ino_data = lcfs_inode_data(ctx, cfs_ino, &cfs_ino_data_buf);
-	if (IS_ERR(cfs_ino_data))
-		return true;
-
-	payload = lcfs_dup_payload_path(ctx, cfs_ino);
+	payload = lcfs_dup_payload_path(test_ctx->ctx, s_ino, 0);
```
alexlarsson, line 45: You can't call both get_backing and dup_payload on the same inode. It needs to check if its a symlink or regular first, and same for get_dir().

fuzzing/main.c
```diff
@@ -26,97 +26,44 @@ int LLVMFuzzerInitialize(int *argc, char ***argv)
...
-	if (test_ctx->recursion_left <= 0)
+	s_ino = lcfs_get_ino_index(test_ctx->ctx, ino, &buffer);
+	if (IS_ERR(s_ino))
 		return true;
 
-	cstr = name;
-	/* Consume the C string.  */
-	while (*cstr)
-		cstr++;
+	lcfs_get_backing(test_ctx->ctx, s_ino, 0, &out_size, &out_path);
```
alexlarsson, line 43: This leaks out_path.

fuzzing/main.c
alexlarsson: Otoh, maybe the callees should verify that and return error in that case.

fuzzing/main.c
giuseppe: thanks, fixed now

fuzzing/main.c
giuseppe: yes, I think it is safer that the calles verify that and do not crash

PR #25, opened 2022-06-01
Add support for fs-verity

alexlarsson:
This adds support for backing content being fs-verity signed, and we verify these against the actual files before using it. Enable by --compute-digest when building the image.

Also we now support `mkcomposefs` (old `writer`) renaming the paths in the image to be by-fsverity-digest, as well as (optionally) copying the content into an object store with such filenames.


alexlarsson:
Forgot to mention, there is also a mount option digest=XXX which verifies that the actual descriptor file has that fs-verity digest.

PR #27, opened 2022-06-07
Move inode flags inline with the inode data and add various flags

alexlarsson:
This makes the inode nr handling simple (its just the offset from the header), and it make the flags more flexible.

Since we now have more space for flags we can also add stuff, like for payloads and xattrs making these save space when missing. We also add a DIGEST_FROM_PAYLOAD flag that avoids storing the digest separately if it is encoded in the payload anyway.


giuseppe (approved):
LGTM

PR #28, opened 2022-06-07
Use a better magic value and embed it in the descriptor header

alexlarsson:
This ia a "more random" value, plus we can use it for sniffing and
quick validation. This replaces the inode_size in the header because
we don't really use that anywmore.


giuseppe (approved):
LGTM

PR #29, opened 2022-06-07
Store payload before inode data

alexlarsson:
This means we can just subtract the payload length from the inode
offset to find it, rather than having to compute the inode
length. This in turn allows us to later grow the inode data
(using new flags) without affecting older kernel implementations,
as they will just ignore the data for flags it doesn't understand.

To find the root inode we now need to store its offset in the
descriptor header.


giuseppe (approved):
LGTM

PR #46, opened 2022-06-09
cfs: Drop the descriptor_path and just use the source option

alexlarsson:
This makes more sense since there can only be one source.


giuseppe (approved):
LGTM

PR #49, opened 2022-06-09
cfs: Validate filenames in directories

alexlarsson:
We disallow embedded nuls or slashes, empty names, ".", and "..".


giuseppe (approved):
LGTM

PR #50, opened 2022-06-10
Fsverity rework

alexlarsson:
This uses a custom fsverity hash implementation to allow us to compute the digest while writing the descriptor. And some cleanups.


giuseppe (approved):
LGTM

PR #51, opened 2022-06-14
kernel: allow multiple basedirs

giuseppe:
allow to specify multiple basedirs -o basedir=/path1:/path2 where
files are looked up sequentilly in a similar way to what overlayfs
does.

### DIFF
--- a/kernel/cfs.c
+++ b/kernel/cfs.c
@@ -35,13 +35,18 @@ MODULE_AUTHOR("Giuseppe Scrivano <gscrivan@redhat.com>");
 
 #include "lcfs-verity.h"
 
+#define CFS_MAX_STACK 500
+
 struct cfs_info {
 	struct lcfs_context_s *lcfs_ctx;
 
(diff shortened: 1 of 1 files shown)


kernel/cfs.c
```diff
@@ -569,6 +638,26 @@ static const struct fs_context_operations cfs_context_ops = {
...
+	for (i = 0; i < fsi->n_bases; i++) {
+		real_file = file_open_root(&(fsi->bases[i]->f_path),
+					   cino->real_path, file->f_flags, 0);
+		if (!IS_ERR(real_file))
+			return real_file;
+
+		last_error = PTR_ERR(real_file);
+	}
+
+	return ERR_PTR(last_error);
```
alexlarsson, line 658: Return first error instead maybe?

alexlarsson: Unless its no-such-file

giuseppe: fixed

PR #53, opened 2022-06-21
Make openssl dependency optional

alexlarsson:
This adds a fallback implementation of sha256, based on the glib code in case openssl is not used. This is primarily useful to make it easy to include the library as a git-submodule. For example, for non-openssl builds of ostree.


giuseppe (approved):
LGTM

PR #58, opened 2022-08-10
Limit in-memory use for inodes

alexlarsson:
This limits the xattr data size to 4k, and splits the dentry data into chunks of (max) 4k which we read in chunks as needed. This drops the in-memory requirement for inodes a lot.

@rhvgoyal does this take care of your issues wrt memory use?

Note: For memory use I will also look at re-using xattr chunks in memory between inodes as they are already shared on disk. Should be rather easy.

### DIFF
--- a/fuzzing/main.c
+++ b/fuzzing/main.c
@@ -29,10 +29,11 @@ bool iter_cb(void *private, const char *name, int namelen, u64 ino, unsigned int
 	struct cfs_xattr_header_s *xattrs;
 	struct cfs_inode_s *s_ino;
 	struct cfs_inode_s buffer;
-	struct cfs_dir_s *dir;
+	struct cfs_dir_data_s dirdata;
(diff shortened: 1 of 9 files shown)


alexlarsson:
Wth is up with armv7:
```
  dump.c:254:24: error: format '%ld' expects argument of type 'long int', but argument 11 has type 'long long unsigned int' [-Werror=format=]
    254 |                 printf("name:%.*s|ino:%" PRIu64
        |                        ^~~~~~~~~~~~~~~~~
  ......
    259 |                        ino.st_size, (uint64_t)0,
        |                                     ~~~~~~~~~~~
        |                                     |
        |                                     long long unsigned int
```

Surely the printf type `PRIu64` (defined in the system header) should match `uint64_t`(also defined in system header).

giuseppe (approved):
LGTM

PR #62, opened 2022-09-20
Improvement in dentry handling and lookup

alexlarsson:
This series converts the dentry chunks into stuff in the vdata section, and then ensures the chunks never straddles pages. We can the use the page cache to look up the chunks without having to do a memcopy. We also implement binary search lookup in the chunks.

Additionally the series has various cleanups and fixes.


giuseppe (approved):
LGTM

PR #63, opened 2022-09-20
Use page cache for xattrs

alexlarsson:
This has some cleanup, and then switches the xattr code to directly use memory from the page cache. This means we don't have to copy this memory for each inode, slimming memory use and sharing memory for caching de-duplicated xattr data.

### DIFF
--- a/kernel/cfs-internals.h
+++ b/kernel/cfs-internals.h
@@ -9,15 +9,6 @@
 #define CFS_MAX_STACK 500
 #define CFS_N_PRELOAD_DIR_CHUNKS 4
 
-struct cfs_buf {
-	struct page *page;
-	void *base;
-};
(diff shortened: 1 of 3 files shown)


kernel/cfs-internals.h
```diff
@@ -9,19 +9,17 @@
...
-	void *base;
-};
-#define CFS_VDATA_BUF_INIT                                                     \
-	{                                                                      \
-		NULL, NULL                                                     \
-	}
-
 struct cfs_inode_data_s {
 	u32 payload_length;
+	char *path_payload; /* Real pathname for files, taget for symlinks */
```
giuseppe, line 14: typo in taget.

giuseppe (approved):
LGTM, just a typo probably not worth another iteration.  Feel free to merge :)

Issue #18, opened 2022-05-25
Implement statfs?


alexlarsson:
In ostree i implemented statfs() to forward e.g. remaining disk space to the backing fs.
I'm not sure if this is the right thing to do or not though, but adding it her for discussions.

PR #71, opened 2022-09-23
kernel: Implement statfs()

alexlarsson:
This retargets statfs() to the first base directory so that we get at least some size info. Free space isn't super important as the fs is not writable, but it makes sense to have at least some value to avoid confusing apps thinging the disk is full.

Fixes https://github.com/containers/composefs/issues/18


giuseppe (approved):
LGTM

PR #72, opened 2022-10-19
cfs: use kern_unmount_array

giuseppe:
prefer kern_unmount_array over kern_unmount since it uses internally synchronize_rcu_expedited that is faster than synchronize_rcu.


alexlarsson:
lgtm

PR #73, opened 2022-11-16
kernel: When opening a file, use open_with_fake_path()

alexlarsson:
This means the looked-up filename will be displayed in e.g. /proc/pid/maps and other places, rather than the backing file.


giuseppe (approved):
LGTM

Issue #74, opened 2022-11-21
Change default behaviour wrt fs-verity checks


alexlarsson:
I keep accidentally running `mkcomposefs --digest-store=path ...` on e.g. systems that don't support fs-verity (or similar situtations), and then get surprised when reading files off the mountpoint fails (with a logged message about wrong/missing fs-verity digest). Then i keep having to look up what the option is to disable verity checks.

I wonder if we should change the behaviour of open, when the inode has a recorded fs-verity digest. What about this set of behaviours:

If the digest=xxx mount option is specified (ie. image itself is verity), then we require:
 * All backing files have an fs-verity digest matching the one in the image
 * The inode in the image specifies a verify digest for the image

Otherwise by default we require:
* If an inode in the image has a verity digest, and the backing file has fs-verity, it need to match

Then we add two new options, one to require all backing files to have a fs-verity of the inode has it, and another one to require all image inodes to have a fs-verity digest. (Bascially one switch for each of the digest=xx requirements). Then we drop the old `noverity` option.

alexlarsson:
(Note: I thing we should never ever allow the backing file to be fs-verity enabled but with a mismatching digest.)

PR #75, opened 2022-11-22
Change when we require fs-verity

alexlarsson:
This changes when we verify fs-verity. The goal is to have it easy to create fs-verity enabled images with `mkcomposefs --digest-store=`, but also make it easy to consume them without requiring you to explicitly disable verity.

The new setup is a single option verity_check=[0,1,2]. 0 means we never require verity. 1 means we veritfy backing file fs-verity if the image specifies a verity digest. 2 additionally means we require the image to specify a digest for all non-empty files.

Fixes https://github.com/containers/composefs/issues/74

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,8 @@ Once it is loaded, it can be used as:
 Mount options:
 
 - `basedir`: is the directory to use as a base when resolving relative content paths.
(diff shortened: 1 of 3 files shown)


giuseppe (approved):
LGTM

PR #76, opened 2022-11-22
Avoid loops when opening basedir files

alexlarsson:
This uses private mounts for the basedirs and do lookup in those. This means the opens in those will never access mounts that are mounted after the mount of the composefs itself, and thus we will never get loops.


giuseppe (approved):
LGTM (after the clang CI is fixed)

PR #77, opened 2022-11-22
Drop the implicit `/` basedir

alexlarsson:
This drops the implicit handling of a rootdir as a basedir, as this behaviour can be unexpected to users. If you really want to use rootfs as basedir you can still explicitly list `/` in the basedir option.


giuseppe (approved):
LGTM

PR #82, opened 2022-11-30
Fix reports from kernel test robot

alexlarsson:
See the kernel robot responses in:
 https://lore.kernel.org/lkml/cover.1669631086.git.alexl@redhat.com/


giuseppe (approved):
LGTM

PR #88, opened 2023-01-10
Various fixes based on kernel list feedback

alexlarsson:
This is mainly based on feedback from @masneyb, but also fixes some recently added checkpatch.pl reports.


giuseppe (approved):
LGTM

PR #91, opened 2023-01-18
Change to a simpler fixed size inode format

alexlarsson:
This follows the discussion with dave chinner on the list. The descriptors are slightly larger (around 40%) but the format, as well as the code is much simpler and more performance.

We also drop the chunking of dirents in favour of using multi-page mappings (using vm_map_ram()) when reading the data from the page cache.,

This also adds various comments describing the file format in the header and generally fixes issues pointed out in the reviews.


giuseppe (approved):
just a nit: LGTM

PR #92, opened 2023-01-18
Correctly handle sizes

alexlarsson:
Use i_size_write to set i_size, and also call inode_set_bytes() to set i_blocks and i_bytes.

Drop the use of simple_get/setattr. These are just confusing as they overwrite i_blocks. The default generic_fillattrs work fine for us.

Also, don't randomly hardcode st_size for dirs, just follow whatever the inode data says, for simplicity.


giuseppe (approved):
LGTM

PR #93, opened 2023-01-19
Fix on-disk struct sizes

alexlarsson:
This puts back the packed attributes on the on-disk struct, because otherwise we get a lot of holes in them, making things larger. We also add some "unused" padding to make the various structures fully or partially cacheline aligned.


giuseppe (approved):
LGTM

PR #99, opened 2023-01-24
kernel: Fix binary search bounds check

alexlarsson:
The binary search was handling the case of looking up something before the table, because when the lookup at start=0 end=0 (mid=0) fails, end was set to mid - 1, which is negative, so still passes the loop check for start_dirent <= end_dirent.


giuseppe (approved):
LGTM

PR #100, opened 2023-01-31
kernel: Simplify cfs_read_iter by using vfs_iter_read()

alexlarsson:
This simplifies the code a bit, and is I believe equivalent.


giuseppe (approved):
LGTM

PR #104, opened 2023-03-02
Initial work on erofs support in mkcomposefs

alexlarsson:
This adds support for --format=erofs to generate an erofs image designed to be used in combination with overlayfs.


giuseppe (approved):
LGTM

PR #108, opened 2023-03-14
Create mount.composefs helper

alexlarsson:
This moves the code to mount erofs+overlayfs images into the libcomposefs so that others can easily use it. It also changes mountcomposefs to /sbin/mount.composefs with a CLI that matches /bin/mount, which allows mount to use it as a helper, like so:

```
$ mount -t composefs -obasedir=objects simple.erofs mnt
```

### DIFF
--- a/libcomposefs/lcfs-mount.c
+++ b/libcomposefs/lcfs-mount.c
@@ -1,0 +1,392 @@
+/* lcfs
+   Copyright (C) 2023 Alexander Larsson <alexl@redhat.com>
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


libcomposefs/lcfs-mount.c
```diff
@@ -0,0 +1,392 @@
...
+
+	mount_flags = 0;
+	if (readonly)
+		mount_flags |= MS_RDONLY;
+
+	res = mount("overlay", state->mountpoint, "overlay", mount_flags,
+		    overlay_options);
+
+fail:
+	free(lowerdir);
```
cgwalters, line 331: Any reason not to cargo cult in the `__attribute__(cleanup)` helpers into this project?

cgwalters (commented):
Superficial/skimmed LGTM

libcomposefs/lcfs-mount.c
alexlarsson: I've considered it, and it would be nice. But never got started on it.

PR #112, opened 2023-03-30
WIP Initial version of cfs-fuse

alexlarsson:
This is a highly efficient implementation of the parts of erofs that we use in composefs.
With this performance of ls -lR of a fuse mount like this is broadly similar to that of a kernel filesystem, except it works rootless. It is stateless in that all it does is mmap the filesystem image and each fuse request is served directly from this mapping.

It does remap the trusted.overlay.* xattrs to user.overlay.*, with the goal of using this with overlayfs, but unfortunately that requires the user of the overlayfs mount option userxattr, which is incompatilble with redirects and metacopy, so this doesn't work atm.

We could perhaps fix that. Alternative approaches is to make it do the reads itself, but then we don't get the page-sharing aspects. One way to fix that is if we ever get the "fuse support to pass fd:s to the caller" kernel patches that has been floating around lately.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -176,6 +176,21 @@ else
 fi
 AM_CONDITIONAL([USE_YAJL],[test "$have_libyajl" = "yes"])
 
+AC_ARG_WITH(fuse,
+        AS_HELP_STRING([--with-fuse], [Build fuse support [default=auto]]),
+        , with_fuse=auto)
+if test "x$with_fuse" != "xno"; then
+  PKG_CHECK_MODULES(FUSE3, [fuse3 >= 3.10.0], [have_fuse3=yes], [have_fuse3=no])
+  if test $have_fuse3 = yes; then
+    AC_DEFINE(HAVE_FUSE3, 1, [Define if fuse3 is available])
+  elif test "x$with_fuse" == "xyes"; then
+    AC_MSG_ERROR([fuse was requested but fuse3 it could not be found])
+  fi
+else
+   have_fuse3=no
+fi
+AM_CONDITIONAL([USE_FUSE3],[test "$have_fuse3" = "yes"])
+
 AC_FUNC_ERROR_AT_LINE
 AC_FUNC_FSEEKO
 AC_HEADER_MAJOR
--- a/tools/Makefile.am
+++ b/tools/Makefile.am
@@ -1,9 +1,13 @@
 bin_PROGRAMS = mkcomposefs
 sbin_PROGRAMS = mount.composefs
-noinst_PROGRAMS = 
+noinst_PROGRAMS =
 
 if USE_YAJL
 noinst_PROGRAMS += composefs-from-json
+endif
+
+if USE_FUSE3
+noinst_PROGRAMS += composefs-fuse
 endif
 
 AM_CFLAGS = $(WARN_CFLAGS) -I$(top_srcdir)/
@@ -16,6 +20,10 @@ mount_composefs_LDADD = ../libcomposefs/libcomposefs.la $(LIBCRYPTO_LIBS)
 
 composefs_from_json_SOURCES = composefs-from-json.c read-file.c read-file.h
 composefs_from_json_LDADD = ../libcomposefs/libcomposefs.la $(LIBS_YAJL) $(LIBCRYPTO_LIBS) $(LIBS_SECCOMP)
+
+composefs_fuse_SOURCES = cfs-fuse.c
+composefs_fuse_LDADD =   ../libcomposefs/libcomposefs.la $(FUSE3_LIBS)
+composefs_fuse_CFLAGS = $(FUSE3_CFLAGS)
 
 TEST_ASSETS_SMALL = \
 	config.json.gz config-with-hard-link.json.gz
--- a/tools/cfs-fuse.c
+++ b/tools/cfs-fuse.c
@@ -1,0 +1,1236 @@
+#define _GNU_SOURCE
+
+#define FUSE_USE_VERSION 34
+
+#include "config.h"
+
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <getopt.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stddef.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
+#include <sys/mount.h>
+#include <sys/types.h>
+#include <linux/limits.h>
+#include <linux/loop.h>
+#include <linux/mount.h>
+#include <linux/fsverity.h>
+#include <fuse_lowlevel.h>
+#include <sys/mman.h>
+
+#include "libcomposefs/lcfs-erofs.h"
+#include "libcomposefs/lcfs-internal.h"
+
+/* TODO:
+ *  Do we want to user ther negative_timeout=T option?
+ */
+
+#define CFS_ENTRY_TIMEOUT 3600.0
+#define CFS_ATTR_TIMEOUT 3600.0
+
+#define ALIGN_TO(_offset, _align_size)                                         \
+	(((_offset) + _align_size - 1) & ~(_align_size - 1))
+
+/* Note: These only do power of 2 */
+#define __round_mask(x, y) ((__typeof__(x))((y)-1))
+#define round_up(x, y) ((((x)-1) | __round_mask(x, y)) + 1)
+#define round_down(x, y) ((x) & ~__round_mask(x, y))
+
+#include "libcomposefs/erofs_fs_wrapper.h"
+
+const uint8_t *erofs_data;
+size_t erofs_data_size;
+uint64_t erofs_root_nid;
+bool erofs_use_acl;
+const struct erofs_super_block *erofs_super;
+const struct lcfs_erofs_header_s *cfs_header;
+const uint8_t *erofs_metadata;
+const uint8_t *erofs_xattrdata;
+uint64_t erofs_build_time;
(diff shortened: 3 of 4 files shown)


alexlarsson:
I updated this to add some code to support reading files (by parsing the overlay redirect xattr). So, with this we have basic support for full access. However, performance may not be great, and also so far we have no verity validation in place.

tools/cfs-fuse.c
```diff
@@ -0,0 +1,1225 @@
...
+	}
+
+	/* We always want the kernel to handle the permissions */
+	fuse_opt_add_arg(&args, "-o");
+	fuse_opt_add_arg(&args, "ro,default_permissions");
+
+	if (fuse_opt_parse(&args, &data, cfs_opts, NULL) == -1)
+		return 1;
+
+	fd = open(data.source, O_RDONLY);
```
cgwalters, line 1142: `O_CLOEXEC`

tools/cfs-fuse.c
```diff
@@ -0,0 +1,1225 @@
...
+}
+
+static uint16_t cfs_inode_datalayout(const erofs_inode *cino)
+{
+	uint16_t i_format = lcfs_u16_from_file(cino->i_format);
+	return (i_format & ((1 << EROFS_I_DATALAYOUT_BITS) - 1)) >>
+	       EROFS_I_DATALAYOUT_BIT;
+}
+
+static bool cfs_inode_is_tailpacked(const erofs_inode *cino)
```
cgwalters, line 132: Kind of here and everywhere...shouldn't we basically `s/cfs_/erofs_`?

cgwalters (commented):
I only gave this a quick skim.  I can imagine this being useful at least for cross-checking, so no objections to me to merging.

configure.ac
```diff
@@ -176,6 +176,8 @@ else
 fi
 AM_CONDITIONAL([USE_YAJL],[test "$have_libyajl" = "yes"])
 
+PKG_CHECK_MODULES(FUSE3, [fuse3 >= 3.10.0])
```
cgwalters, line 179: But, let's make this a build-time option and not a hard requirement.

tools/Makefile.am
```diff
@@ -3,7 +3,7 @@ sbin_PROGRAMS = mount.composefs
 noinst_PROGRAMS = 
 
 if USE_YAJL
-noinst_PROGRAMS += composefs-from-json
+noinst_PROGRAMS += composefs-from-json cfsfuse
```
cgwalters, line 6: Naming schemes for filesystem tools is messy, but `cfsfuse` feels...well, I guess it's accurate, just kind of terse.  `mount.cfsfuse` maybe?  Dunno.

cgwalters (changes requested):

cgwalters:
@giuseppe asked:
> how does it compare to the fuse file system that already exists for erofs?  Could this be handled in erofs-tools?

I think a key part of the answer here is that we'd have to add support for the "basedir" bits; it looks to me like the [erofs-utils fuse implementation](https://github.com/hsiangkao/erofs-utils/blob/dev/fuse/main.c) is really really basic.  It doesn't have `overlay.redirect` support  I think.

But yes...hmm, maybe it is actually simpler to see if they'd take a patch for that; or worst case if not, just fork the binary for now and add the support.  That'd make it easier to unfork later hopefully.

Oh and, it's still using the old FUSE API.  It also looks like it's not multithreaded?

alexlarsson:
> @giuseppe asked:
> 
> > how does it compare to the fuse file system that already exists for erofs?  Could this be handled in erofs-tools?
> 
> I think a key part of the answer here is that we'd have to add support for the "basedir" bits; it looks to me like the [erofs-utils fuse implementation](https://github.com/hsiangkao/erofs-utils/blob/dev/fuse/main.c) is really really basic. It doesn't have `overlay.redirect` support I think.
> 
> But yes...hmm, maybe it is actually simpler to see if they'd take a patch for that; or worst case if not, just fork the binary for now and add the support. That'd make it easier to unfork later hopefully.
> 
> Oh and, it's still using the old FUSE API. It also looks like it's not multithreaded?

There are two replies here.

The first one is that the cfsfuse implements the combination of erofs+overlayfs (although strictly limited to the features composefs uses), whereas erofs fuse supports only the erofs part (but with much more erofs features). It is not possible to use a non-root overlayfs with  erofs-fuse to implement the composefs features, because we fundamentally rely on the overlayfs redirect feature, which is incompatible with the overlay userxattrs option which would let us use the redirects.

The second reply is that the two are implemented very differently. The erofs one uses the generic implementation from the erofs-utils, and that code has all sorts of generic code needed for all the tools in that codebase. However, the cfsfuse implementation has inline support for the just the required features. This allows the cfsfuse implementation to be extremely tight. For example, there are zero allocations at runtime, every fs operation is just a few dereferences into a mmaped copy of the filesystem data.

The end result is that cfsfuse performs much better (for the subset we need):

```
$ erofsfuse cs9.cfs m
$ time ls -lR m > /dev/null
real	0m5,149s
user	0m0,275s
sys	0m0,723s
$ time ls -lR m > /dev/null
real	0m4,939s
user	0m0,247s
sys	0m0,654s
```
vs
```
$ tools/cfsfuse m/ -o source=cs9.cfs,basedir=objects
$ time ls -lR m > /dev/null
real	0m1,100s
user	0m0,192s
sys	0m0,538s
$ time ls -lR m > /dev/null
real	0m0,530s
user	0m0,155s
sys	0m0,318s
```

tools/cfs-fuse.c
alexlarsson: Not everywhere. This is cfs-fuse, so i think it makes sense to use the cfs_ prefix for it.

However, this particular function, and e few others are very erofs specific, so I guess it makes sense to have them be erofs_...

tools/Makefile.am
alexlarsson: `mount.*` is kinda reserved for /bin/mount helpers.

Other things seem be nonstandard, but mostly `xxx-fuse`:
* erofsfuse
* fuse2fs
* fuse-overlayfs
* rofiles-fuse
* vmhgfs-fuse   
* vmware-vmblock-fuse
* gvfsd-fuse 
* revokefs-fuse
* zfs-fuse-scrub
* zfs-fuse

I think i'll go with `composefs-fuse`, to match the existing `mount.composefs` and `mkcomposefs` tools.

alexlarsson:
I pushed some further tweaks that should improve performance more (use splice, use clone-fd).

tools/cfs-fuse.c
```diff
@@ -0,0 +1,1229 @@
...
+
+	tailpacked = erofs_inode_is_tailpacked(cino);
+	if (!tailpacked) {
+		fuse_reply_err(req, EINVAL);
+		return;
+	}
+
+	const uint8_t *inline_data = ((uint8_t *)cino) + isize + xattr_size;
+
+	memcpy(name_buf, inline_data, file_size);
```
giuseppe, line 657: same here, do we need to make sure we don't write after the buffer?

giuseppe (commented):
ok then I am fine to merge, just some comments

tools/cfs-fuse.c
alexlarsson: yeah, added

hsiangkao:
> > @giuseppe asked:
> > > how does it compare to the fuse file system that already exists for erofs?  Could this be handled in erofs-tools?
> > 
> > 
> > I think a key part of the answer here is that we'd have to add support for the "basedir" bits; it looks to me like the [erofs-utils fuse implementation](https://github.com/hsiangkao/erofs-utils/blob/dev/fuse/main.c) is really really basic. It doesn't have `overlay.redirect` support I think.
> > But yes...hmm, maybe it is actually simpler to see if they'd take a patch for that; or worst case if not, just fork the binary for now and add the support. That'd make it easier to unfork later hopefully.
> > Oh and, it's still using the old FUSE API. It also looks like it's not multithreaded?
> 
> There are two replies here.
> 
> The first one is that the cfsfuse implements the combination of erofs+overlayfs (although strictly limited to the features composefs uses), whereas erofs fuse supports only the erofs part (but with much more erofs features). It is not possible to use a non-root overlayfs with erofs-fuse to implement the composefs features, because we fundamentally rely on the overlayfs redirect feature, which is incompatible with the overlay userxattrs option which would let us use the redirects.
> 
> The second reply is that the two are implemented very differently. The erofs one uses the generic implementation from the erofs-utils, and that code has all sorts of generic code needed for all the tools in that codebase. However, the cfsfuse implementation has inline support for the just the required features. This allows the cfsfuse implementation to be extremely tight. For example, there are zero allocations at runtime, every fs operation is just a few dereferences into a mmaped copy of the filesystem data.
> 
> The end result is that cfsfuse performs much better (for the subset we need):
> 
> ```
> $ erofsfuse cs9.cfs m
> $ time ls -lR m > /dev/null
> real	0m5,149s
> user	0m0,275s
> sys	0m0,723s
> $ time ls -lR m > /dev/null
> real	0m4,939s
> user	0m0,247s
> sys	0m0,654s
> ```
> 
> vs
> 
> ```
> $ tools/cfsfuse m/ -o source=cs9.cfs,basedir=objects
> $ time ls -lR m > /dev/null
> real	0m1,100s
> user	0m0,192s
> sys	0m0,538s
> $ time ls -lR m > /dev/null
> real	0m0,530s
> user	0m0,155s
> sys	0m0,318s
> ```

Yes， I suggest don't use erofsfuse for now (or if you later have extended features) since erofsfuse is a too native implementation. I asked some school students to work on this this summer to improve it but erofs-utils won't address composefs use cases since I think it should be in composefs project.

PR #114, opened 2023-04-20
Library cleanup

alexlarsson:
* Only export public ELF symbols
* Split lcfs-writer.c into several files
* Add some basic tests
* Add optional support for valgrinding tests


giuseppe (approved):
LGTM

PR #115, opened 2023-04-21
Various packaging and distribution fixes

alexlarsson:
* Relicense lib to LGPL
* Fix make dist
* Add spec file
* Add pkg-config file


giuseppe (approved):
LGTM

PR #116, opened 2023-04-21
Fix pkg-config dependencies

alexlarsson:
Actually the library doesn't need libyajl or libfsverity, those are only used by the tools, so fix the .pc file.


giuseppe (approved):
LGTM

PR #119, opened 2023-04-24
Drop libsfverity dependency

alexlarsson:
We don't currently use it other than to enable fs-verity, which really is just a single ioctl call. So, lets call that directly instead.


giuseppe (approved):
LGTM

PR #127, opened 2023-05-29
a bunch of warnings fixes

giuseppe:
the only breaking change is for `lcfs_build`


alexlarsson:
LGTM, but fails the clang formatter.

PR #138, opened 2023-06-07
Drop kernel implementation for now

alexlarsson:
This simplifies the userspace and docs by focusing on the current main format, and it also means we won't be constantly exposing a kernel module code that is not currently being worked on, and which likely will go out of sync with upstream linux quickly.


giuseppe (approved):
LGTM

PR #140, opened 2023-06-07
Update test checksums and ensure failure is detected

alexlarsson:
It turns out that a commit (0201723103acdf3f243b2f45d0c8da5e75c97c7e) we did changed the checksums of some tests, but we didn't update the checksums, so update those.

Also, ensure the tests actually fail in case of such mismatched checksums.


alexlarsson:
Yeah, we're looking at this. Its caused by some timezone issues in the time parsing code for the json files.

cgwalters:
Note the failing test is on s390x.  Some sort of endianness issue in erofs?

hsiangkao:
> Note the failing test is on s390x. Some sort of endianness issue in erofs?

I don't have big-endian environment  on my side, I could use QEMU to debug later.
Is there some images available so I could try?

alexlarsson:
I doubt it is an endianness problem, I think only the json specific images fail.
Lemme temporarily disable the commit that makes the test fails due to the checksums mismatching and we can see which arches and images fail.

alexlarsson:
Interestingly all arches passed except the s390x one, and that one failed to reproduce the image checksums for *all* images, not just the json-derived ones. I guess that means the json timezone parsing issue really is fixed, but we have some big-endian issue too.

alexlarsson:
This is weird. I pushed a new commit to the branch, but its not appearing here.

tools/composefs-from-json.c
```diff
@@ -667,6 +673,8 @@ int main(int argc, char **argv)
 	cleanup_free FILE **input_files = NULL;
 	bool no_sandbox = false;
 
+	tzset();
```
giuseppe, line 676: shouldn't we enforce `TZ=GMT` before?

alexlarsson: I don't think we need that, as we use timegm(). I'm honestly not sure we need tzset() either...

giuseppe (approved):
LGTM

PR #143, opened 2023-06-08
mkcomposefs: Use FICLONE if available for objects

cgwalters:
This makes generating an object store a lot more efficient if available; one can construct a filesystem tree however, run `mkcomposefs --digest-store=` on it, and if the digest store and the source tree are on the same filesystem, then we get O(1) copies.

One interesting twist with this is the intersection of reflinks and fsverity; the only Linux filesystem supporting both today is btrfs, and I gave this a quick test there and it seems to work. (There are pending xfs patches for fsverity which would also
 enable this there)


alexlarsson (approved):
lgtm

PR #146, opened 2023-06-12
mount: Use the new syntax for overlayfs datalower dirs

alexlarsson:
The final merged variant of datalowers uses double-colon for every datalower directory, not just the first. I.e:

 -o lower1:lower2::lowerdata1::lowerdata2.

### DIFF
--- a/libcomposefs/lcfs-mount.c
+++ b/libcomposefs/lcfs-mount.c
@@ -364,7 +364,7 @@ static char *compute_lower(const char *imagemount,
 	escape_mount_option_to(imagemount, lower);
 
 	for (i = 0; i < state->options->n_objdirs; i++) {
-		if (with_datalower && i == 0)
+		if (with_datalower)
 			strcat(lower, "::");
 		else
 			strcat(lower, ":");


cgwalters:
Is there a reference for this?  I searched https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt and the latest fs/overlayfs/super.c code in vain for anything related to `::`.

cgwalters:
And now I'm confused by the fallback handling in...oh wow, I see, we just try the mount with the `::` and retry with `:` if that fails.  Hmm.  For one thing we probably want to add a compile time flag for this for cases where we know the new kernel API exists.  We can probably make that option default to *off* for now, and people developing the "new composefs" can turn it on?

Anyways, seems sane to me...

cgwalters (approved):

alexlarsson:
I have the workaround atm, because otherwise composefs doesn't work on any upstream kernel. At some point we may want to drop that, either completely or as an option.

alexlarsson:
@cgwalters Also, if we can rely on `::` support we know the kernel supports data only lower dirs, which means that we can skip generating a bunch of whiteouts that we currently do in ostree. I think that is a decision we have to make eventually, i.e. whether we want to support older kernels with the images we generate.

PR #148, opened 2023-06-12
gitignore: Add `Makefile-*.am.inc`

cgwalters:
ostree uses composefs as a git submodule today; this is what we do in libglnx too.


giuseppe (approved):
LGTM

PR #156, opened 2023-06-16
mkcomposefs: Add --print-digest-only

cgwalters:
In some cases, it's very useful to just see what the composefs digest of an input tree is, without actually writing an image.  One could do this externally by allocating a tempfile, but it's nicer for us to do internally.

Here we allocate an anonymous `O_TMPFILE` and hence none of the written data will ever persist to disk.


cgwalters:
Ah now that I dig farther, we actually support not having a file set at all, because we do all the hashing outside of writing.  Cool.

giuseppe (approved):
LGTM

PR #157, opened 2023-06-19
lcfs: Drop function to remove child

cgwalters:
In general our use case is really constructing an entire filesystem
tree in one shot from external data; we don't have a use case for
dynamic mutations, which is what this function does.

---


giuseppe:
I am fine to remove it now, since as you said we don't have yet a user, but I think in the future it can be useful to merge multiple layers in a container image into a single one.

alexlarsson:
Yeah, this was initially used in ostree to apply a whiteout layer. But, I'm fine with removing it for now.

alexlarsson (approved):
lgtm

PR #159, opened 2023-06-21
lib: Don't `qsort(NULL)`

cgwalters:
This is UB, though I think glibc is resilient to it. Detected by running with `-fsanitizer=undefined`.


giuseppe (approved):
LGTM

PR #161, opened 2023-06-21
ci: Enable sanitiziers by default

cgwalters:
Depends:

 - #160 
 - #159 

These are cheap enough and valuable enough that we should really run with them on by default in CI.


alexlarsson (approved):
LGTM, but needs a rebase now that the dependencies are merged

PR #162, opened 2023-06-21
lib: Avoid O(N^2) behavior on cleanup

cgwalters:
lib: Use common function to remove all children

Prep for further cleanup.

---

lib: Avoid O(N^2) behavior on cleanup

This library was written in a very general way with an API
`lcfs_node_remove_child` that required removing an arbitrary
member of the children array.  I found a bug in that code
previously in that it should have been using `memmove`, not
`memcpy`.

However, I think there's no need for us to support arbitrary
tree computations.

Today at process exit we walk over the in-memory filesystem
and we end up with an O(N^2) behavior because we keep removing
the first child.

Instead, because we *only* need to care about removing all
children at once, just iterate over them and then free them
as we go.

Note: I only noticed this code because of the `memmove` fix, not
because I saw some performance problem in practice.  It turns
out computers are fast...

---

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -754,25 +754,6 @@ void lcfs_node_make_hardlink(struct lcfs_node_s *node, struct lcfs_node_s *targe
 	target->inode.st_nlink++;
 }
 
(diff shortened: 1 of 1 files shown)


giuseppe (approved):
LGTM

PR #163, opened 2023-06-21
Drop support for ctime

cgwalters:
As discussed in https://github.com/containers/storage/pull/1646 there is no support in EROFS for ctime != mtime.  The current code does nothing with it.

Drop support for it in the API.


giuseppe (approved):
LGTM

PR #166, opened 2023-07-04
composefs-from-json: skip empty dir components

giuseppe:
solves a problem where a path like "sys/" would cause to add a node with an empty name.


alexlarsson (approved):
Need formating, but other LGTM

giuseppe:
CI is green

cgwalters (approved):

PR #173, opened 2023-08-07
tools: Drop ostree-convert-commit.py

cgwalters:
OSTree upstream now links to libcomposefs and can
synthesize cfs images directly.


alexlarsson (approved):
lgtm

PR #174, opened 2023-08-13
ci: add riscv64 support

michalbiesek:


alexlarsson:
Is there any particular reason to run CI on riscv64? Like, does it test some codepath that otherwise would not get tested, such as big endian?

Very little in this code is arch dependent, so if this is just gonna cause CI to be slower, or more likely to fail due to some risc CI machine limit, I don't quite see the point of this.

Issue #172, opened 2023-07-21
handle nested whiteouts


cgwalters:
See https://github.com/ostreedev/ostree/issues/2712

Basically...we're now shipping support in ostree for embedding whiteouts.  This allows c/image (podman etc.) to be directly pointed at this alternative root in a read-only fashion.  Container images shipped this way are "lifecycle bound" with the host (and gain benefits of dedup actually and the efficient ostree on-the-wire deltas).

But this only works because ostree itself doesn't use overlayfs (ostree actually predates overlayfs).

In a composefs future, because overlayfs doesn't nest, we're going to need to figure out how to handle this fun special case.

In a [unified storage world](https://github.com/containers/composefs/issues/125) things are inherently better here, but hard requring that would actually be an "API break".  

I guess another way to say this is that everyone [turning on the ostree composefs support](https://ostreedev.github.io/ostree/composefs/) is just going to break if they have nested containers today.

cgwalters:
So...half baked thoughts.  We can clearly detect this situation in ostree when we're generating the composefs image.  I think probably what we need to do here is fall back to continuing to create that sub-portion of the filesystem tree as hardlinks outside of a composefs mount (which we already do today actually).  Then, we set up a dynamic bind mount from there to that hardlinked subtree.

I think that'd work...

alexlarsson:
I don't think this is the only issue we're going to have, is it? I mean, you also can't store overlayfs xattrs in the composefs image.

@amir73il Is there a way to "recurse" overlayfs dirs for this? I.e. have an overlayfs mount that looks like it contains files such as whiteouts and regular files with overlayfs xattrs set on them?

alexlarsson:
I wonder actually if we should have libcomposefs error out if the source data specifies such files. Currently they will produce pretty broken resulting images.

amir73il:
Generally, overlayfs mount doesn't accept an overlayfs as the upper/work dir.
There are scattered assumptions about upper/work fs not being "wierd" and I don't know for sure what all these assumptions are. Need to consult with Miklos if this is important enough.
Note that the requirement that upper/work fs are not "remote" have already been relaxed to allow to support fuse as upper fs.

alexlarsson:
And same with whiteouts, we would like a way to store a file in a lower that overlayfs doesn't consider a whiteout, but looks like a whiteout in the resulting overlayfs mountpoint.

amir73il:
[...]

PR #175, opened 2023-08-17
erofs: Escape overlayfs features

alexlarsson:
This escapes overlayfs xattrs and whiteouts according to the support in this patch series:
  https://lore.kernel.org/linux-unionfs/cover.1692198910.git.alexl@redhat.com/

Note: We should probably wait with merging this until we have more feedback on the kernel list about it.

Fixes  https://github.com/containers/composefs/issues/172

### DIFF
--- a/libcomposefs/lcfs-writer-erofs.c
+++ b/libcomposefs/lcfs-writer-erofs.c
@@ -19,6 +19,7 @@
 #include "config.h"
 
 #include "lcfs-internal.h"
+#include "lcfs-utils.h"
 #include "lcfs-writer.h"
 #include "lcfs-fsverity.h"
 #include "lcfs-erofs.h"
@@ -33,6 +34,7 @@
 #include <dirent.h>
 #include <sys/xattr.h>
 #include <sys/param.h>
+#include <sys/sysmacros.h>
 #include <assert.h>
 #include <linux/fsverity.h>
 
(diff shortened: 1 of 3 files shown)


cgwalters:
> Note: We should probably wait with merging this until we have more feedback on the kernel list about it.

Yeah makes sense; moved to draft for that reason

alexlarsson:
Patch series 4 is more or less acked, so lets get this in

libcomposefs/lcfs-writer-erofs.c
```diff
@@ -863,11 +865,44 @@ static int write_erofs_shared_xattrs(struct lcfs_ctx_s *ctx)
...
 	int ret;
 
-	if ((node->inode.st_mode & S_IFMT) == S_IFREG && node->inode.st_size > 0) {
+	/* First escape all existing "trusted.overlay.*" xattrs */
+	for (size_t i = 0; i < lcfs_node_get_n_xattr(node); i++) {
+		const char *name = lcfs_node_get_xattr_name(node, i);
+
+		if (str_has_prefix(name, "trusted.overlay.")) {
+			cleanup_free char *renamed =
+				str_join("trusted.overlay.overlay.",
```
cgwalters, line 891: Is it really worth carrying `str_join` over just `asprintf ("trusted.overlay.overlay.%s", name + strlen("trusted.overlay."))` ?

libcomposefs/lcfs-writer-erofs.c
```diff
@@ -899,6 +934,24 @@ static int add_overlayfs_xattrs(struct lcfs_node_s *node)
 		}
 	}
 
+	/* escape whiteouts */
+	if (type == S_IFCHR && node->inode.st_rdev == makedev(0, 0)) {
+		struct lcfs_node_s *parent = lcfs_node_get_parent(node);
+
+		lcfs_node_set_mode(node,
+				   S_IFREG | (lcfs_node_get_mode(node) & ~S_IFMT));
+		ret = lcfs_node_set_xattr(node, "trusted.overlay.overlay.whiteout",
```
cgwalters, line 943: Optional future cleanup: we could have a `linux-overlayfs-internals.h` header that includes `#define`s that link to the upstream source code.

cgwalters (approved):
Looks sane, but I didn't sanity check this versus the kernel APIs.

Also, we will clearly want some integration tests for this in the future.

alexlarsson:
Yeah, I plan to land all the outstanding stuff, and then do work on adding more tests in general.

libcomposefs/lcfs-writer-erofs.c
alexlarsson: I might look at extracting this. Although we can't use the kernel headers here, the way the xattr defines are structured in the kernel is a bit weird.

Issue #167, opened 2023-07-05
Add bloom filter data to erofs images


alexlarsson:
This patch-set adds support to erofs for bloom filters to speed up negative xattr lookups:
https://lists.ozlabs.org/pipermail/linux-erofs/2023-July/008565.html

This change is backwards compat in the sense that we could add these to the erofs images we generate and old kernels would still work with the images. It would be good to get this into our erofs image format before we formally lock down the v0 binary format.

hsiangkao:
Cc @lostjeffle

alexlarsson:
v3 was posted here: https://lists.ozlabs.org/pipermail/linux-erofs/2023-July/008669.html

cgwalters:
> It would be good to get this into our erofs image format before we formally lock down the v0 binary format.

Is that the *only* thing you see blocking a v0?  

I feel like we should do some sort of group "audit" or pass-through of *exactly* what is in the EROFS images we're generating and whether it covers everything it needs to and it looks sane or so.  Did anyone already do that?

alexlarsson:
Offhand the other thing I see is whether to keep the `00` to `ff` whiteouts, which are only needed if the new overlayfs data-only layers are unsupported.

alexlarsson:
Well, the other thing is that I'd like to see the overlayfs fs-verity digest xattr support in a released kernel. But we're *almost* there with that.

cgwalters:
> Offhand the other thing I see is whether to keep the 00 to ff whiteouts, which are only needed if the new overlayfs data-only layers are unsupported.

Ah yes.  Related to that Giuseppe suggested somewhere to at least short term move them to a `.cfs` subdirectory, that seems a little nicer - then we only need one whiteout.

I dunno.  Hmmm....

Will existing erofs.ko barf if we have the bloom filters now?  Or we just want to be sure it's landed upstream before committing to it?

hsiangkao:
> Or we just want to be sure it's landed upstream before committing to it?

xattr bloom filter on-disk format is almost fixed (I have see no more problem), I think it will be landed upstream when the next merge window is open.

alexlarsson:
Adding the xattrs is compat, we just want to wait until the erofs format changes are deemed final.

alexlarsson:
> > Or we just want to be sure it's landed upstream before committing to it?
> 
> xattr bloom filter on-disk format [1] is almost fixed (I have see no more problem), I think it will be landed upstream when the next merge window is open.
> 
[...]

PR #176, opened 2023-08-21
Erofs bloom filter

alexlarsson:
This adds the COMPAT_XATTR_FILTER feature flag, and computes h_name_filter in all the inode xattr header.

Fixes https://github.com/containers/composefs/issues/167

This is based on and tested against the code in the [dev branch of erofs.git](https://git.kernel.org/pub/scm/linux/kernel/git/xiang/erofs.git/log/?h=dev), so this is not upstream quite yet. However, this likely will go in unchanged.

@hsiangkao wanna have a quick look?


alexlarsson:
NOTE: This is a format-breaking change, so its also updates the test checksums. We should merge this before freezing the format.

libcomposefs/erofs_fs.h
```diff
@@ -127,11 +128,30 @@ static inline bool erofs_inode_is_data_compressed(unsigned int datamode)
...
+#define EROFS_INODE_LAYOUT_COMPACT	0
+/* 64-byte on-disk inode */
+#define EROFS_INODE_LAYOUT_EXTENDED	1
+
 struct erofs_inode_chunk_info {
 	__le16 format;		/* chunk blkbits, etc. */
 	__le16 reserved;
 };
 
+union erofs_inode_i_u {
```
cgwalters, line 141: Seems like some of this could be split out as a prep patch?

libcomposefs/lcfs-writer-erofs.c
```diff
@@ -36,6 +36,121 @@
 #include <assert.h>
 #include <linux/fsverity.h>
 
+/* The xxh32 hash function is copied from xxhash with this copyright:
```
cgwalters, line 39: From *where* precisely?  A git URL and exact digest would be useful.

Taking the copy from the Linux kernel and massaging it to compile may not actually be *too* hard...

libcomposefs/erofs_fs.h
alexlarsson: Well, actually, it does tweak the code to work with some new naming in the new file, but its nice to merge those to have each commit build.

libcomposefs/lcfs-writer-erofs.c
alexlarsson: I took it from:
https://www.spinics.net/lists/kernel/msg4859620.html
Which is taken from the kernel, but initially was from 
static const uint32_t PRIME32_1 = 2654435761U;

alexlarsson: https://github.com/torvalds/linux/blob/master/lib/xxhash.c

cgwalters: OK makes sense, but I more meant let's add this data into the comment in the code, so someone knows where to look for updates.

hsiangkao:
> @hsiangkao wanna have a quick look?

Sorry for a bit late reply since I'm busy in the last test for commits of the next cycle.
Overall it looks good to me, will upstream xattr bloom filter feature once the merge window opens.

alexlarsson:
Added the link to the kernel source of xxh32()

libcomposefs/erofs_fs.h
alexlarsson: The changes were proposed upstream today, so I think we can merge the whole thing soon.

alexlarsson:
upstreaming is here: https://lists.ozlabs.org/pipermail/linux-erofs/2023-August/009427.html

hsiangkao:
> upstreaming is here: https://lists.ozlabs.org/pipermail/linux-erofs/2023-August/009427.html

it already lands upstream.

giuseppe (approved):
LGTM

PR #178, opened 2023-08-25
A few smaller/more obvious fixes from running `clang-analyzer`

cgwalters:
writer: Drop dead assignment

Spotted by clang-analyzer.

---

mkcomposefs: Silence an unused variable warning

clang-analyzer complains about this.  While we're here, make
the variable `const`.

---

mkcomposefs: Switch to declare-and-initialize to silence clang-analyzer

clang-analyzer complains about a leak; I think it's wrong, but changing
things to atomically initialize the variables silences it.

---

mkcomposefs: Fix possibly garbage return in error path

Found by clang-analyzer.

(At this point, I have to say: I don't understand why in 2021 we
 created a new project in the C programming language...)

---


cgwalters:
(One backing motivation here is we now build composefs in ostree, and that gets run through more static analyzers)

alexlarsson (approved):
lgtm

PR #179, opened 2023-08-25
mount: Fix memory leak in error path

cgwalters:
Found by clang-analyzer.  This one is really a textbook example of why the cleanup attribute solves so many bugs.


cgwalters:
This one was more subtle, so splitting it out to a separate PR.  Only compile tested locally, but our integration test will cover at least the fallback path I believe.

alexlarsson (approved):
lgtm

PR #180, opened 2023-08-28
writer: Fix missing error return value checks

eriksjolund:
I tried to fix some missing error return value checks.


alexlarsson (approved):
lgtm

PR #181, opened 2023-08-29
Use cleanup_fdp to quiet clang-analyzer

cgwalters:
This avoids a "dead nested assignment" warning because clang doesn't understand that we're explicitly setting to -1 for the cleanup function.


alexlarsson (commented):
lgtm

alexlarsson (approved):

Issue #182, opened 2023-08-31
support for reading .cfs images


jluebbe:
For some use-cases, it would be useful to have an API to recreate an `lcfs_node` tree from `.cfs` image:

- fetching objects needed by an image from a remote object store
- removing unused objects from a local, shared object store given a set of `.cfs` images
- listing missing objects given a `.cfs` image

My main motivation for this would be RAUC's [artifact updates](https://github.com/rauc/rauc/issues/969), where we could use this to stream missing objects from the remote update bundle.

alexlarsson:
In main now

PR #185, opened 2023-09-08
Add support for decoding erofs into lcfs_node, and some tooling using it

alexlarsson:
This will help with fixing
https://github.com/containers/composefs/issues/182


alexlarsson:
I updated the composefs-info escaping to be a bit more strict. 

The dump format is very simple and readable, yet efficiently streamable, I wonder if that is a good format for streaming data to mkcomposefs. Its maybe not super extensible though...

libcomposefs/lcfs-writer-erofs.c
```diff
@@ -1303,3 +1304,512 @@ int lcfs_write_erofs_to(struct lcfs_ctx_s *ctx)
...
+
+static bool erofs_inode_is_tailpacked(const erofs_inode *cino)
+{
+	return erofs_inode_datalayout(cino) == EROFS_INODE_FLAT_INLINE;
+}
+
+static struct lcfs_node_s *lcfs_build_node_from_image(struct lcfs_image_data *data,
+						      uint64_t nid);
+
+static int erofs_readdir_block(struct lcfs_image_data *data,
```
cgwalters, line 1367: It looks like this is a copy-paste-edit of the relevant code from cfs-fuse.c?  Might be nice to at least add a comment indicating that if one is changed, to change the other.

Or, to at least figure out how we can share more code.

libcomposefs/lcfs-writer-erofs.c
```diff
@@ -1303,3 +1304,512 @@ int lcfs_write_erofs_to(struct lcfs_ctx_s *ctx)
...
+		if (lcfs_node_add_child(parent, child, /* Takes ownership on success */
+					name_buf) < 0)
+			return -1;
+		steal_pointer(&child);
+	}
+
+	return 0;
+}
+
+const char *erofs_xattr_prefixes[] = {
```
cgwalters, line 1426: Surely it'd be trivial to share this in a `lcfs-internal-erofs.h` with the fuse code?

tools/composefs-info.c
```diff
@@ -0,0 +1,295 @@
...
+	command = argv[1];
+
+	if (argc <= 2) {
+		fprintf(stderr, "No image path specified\n");
+		usage(bin);
+		exit(1);
+	}
+	image_path = argv[2];
+
+	fd = open(image_path, O_RDONLY);
```
cgwalters, line 272: `| O_CLOEXEC`

tools/composefs-info.c
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	if (strcmp(command, "ls") == 0) {
+		print_node(root, "");
+	} else if (strcmp(command, "dump") == 0) {
+		dump_node(root, "");
+	} else if (strcmp(command, "objects") == 0) {
+		print_objects(root);
+	} else {
+		fprintf(stderr, "Unknown command '%s'\n", command);
```
cgwalters, line 289: This should also be `errx` right?

tools/composefs-info.c
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	lcfs_node_unref(root);
```
cgwalters, line 292: opt: This could use the cleanup attribute.

tools/composefs-dump.c
```diff
@@ -0,0 +1,94 @@
...
+	src_path = argv[1];
+
+	if (argc <= 2) {
+		fprintf(stderr, "No destination path specified\n");
+		usage(bin);
+		exit(1);
+	}
+	dst_path = argv[2];
+
+	fd = open(src_path, O_RDONLY);
```
cgwalters, line 67: `| O_CLOEXEC`

libcomposefs/lcfs-writer-erofs.c
alexlarsson: They are note quite the same because the fuse one is driven by the fuse callback apis, but yeah, we should try to share a bit more.

libcomposefs/lcfs-writer-erofs.c
alexlarsson: yeah

alexlarsson:
Rebased on latest and fixed some minor issues. This is still not ready.

alexlarsson:
TODO:

- [ ] Make cfs-fuse and image-to-node handle the new inlined files
- [ ] Share more code between cfs-fuse and image-to-node
- [ ] Add some more testing

alexlarsson:
I rebased on master, implemented missing features in fuse, generally cleaned up the code and added a few more tests.

I think with this in we're ready to freeze the on-disk format.

Once this is in I will focus on adding even more tests.

libcomposefs/lcfs-erofs-internal.h
```diff
@@ -0,0 +1,134 @@
...
+	res = malloc(prefix_len + name_len + 1);
+	if (res == NULL) {
+		errno = ENOMEM;
+		return NULL;
+	}
+	memcpy(res, prefix, prefix_len);
+	memcpy(res + prefix_len, name, name_len);
+	res[prefix_len + name_len] = 0;
+
+	return res;
```
cgwalters, line 131: This is all just `return asprintf("%s%.*s", prefix, name_len, name)` right?

libcomposefs/lcfs-utils.h
```diff
@@ -19,12 +19,41 @@
...
+	char *s2 = malloc(len);
+	if (s2 == NULL) {
+		errno = ENOMEM;
+		return NULL;
+	}
+	memcpy(s2, s, len);
+	return s2;
+}
+
+static inline char *str_join(const char *a, const char *b)
```
cgwalters, line 45: optional: This is just code motion, so OK, but same question re `asprintf` - can be a followup.

libcomposefs/lcfs-internal.h
```diff
@@ -29,6 +29,28 @@
  */
 #define LCFS_BUILD_INLINE_FILE_SIZE_LIMIT 64
 
+#define OVERLAY_XATTR_USER_PREFIX "user."
```
cgwalters, line 32: Aweseome, I think this helps a lot to have these.  Then we can also followup with some links to the kernel sources.

libcomposefs/lcfs-writer-erofs.c
```diff
@@ -1116,12 +1116,20 @@ static int add_overlayfs_xattrs(struct lcfs_node_s *node)
 					  "", 0);
 		if (ret < 0)
 			return ret;
+		ret = lcfs_node_set_xattr(node, OVERLAY_XATTR_USERXATTR_WHITEOUT,
```
cgwalters, line 1125: Hum, interesting...but can you help me understand this?  Do we need the userxattrs for unprivileged overlayfs mounts?  This seems tricky because it's redundant data in the privileged case, no?  Or maybe it isn't much data and doesn't matter.

libcomposefs/lcfs-writer.c
```diff
@@ -628,6 +629,40 @@ struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
...
+	image_data_size = s.st_size;
+
+	image_data = mmap(0, image_data_size, PROT_READ, MAP_PRIVATE, fd, 0);
+	if (image_data == MAP_FAILED) {
+		return NULL;
+	}
+
+	node = lcfs_load_node_from_image(image_data, image_data_size);
+	if (node == NULL) {
+		errsv = errno;
```
cgwalters, line 655: We have a handy `PROTECT_ERRNO` macro btw.

tools/composefs-info.c
```diff
@@ -0,0 +1,294 @@
...
+		}
+		printf("\n");
+
+		print_node(child, path);
+	}
+}
+
+static void digest_to_string(const uint8_t *csum, char *buf)
+{
+	static const char hexchars[] = "0123456789abcdef";
```
cgwalters, line 123: As far as I can tell, this would literally be the *fourth* copy of this in tree...

tools/composefs-info.c
```diff
@@ -0,0 +1,294 @@
...
+static void print_objects(struct lcfs_node_s *node)
+{
+	Hash_table *ht = hash_initialize(0, NULL, str_ht_hash, str_ht_eq, NULL);
+	if (ht == NULL)
+		errx(EXIT_FAILURE, "Out of memory");
+
+	get_objects(node, ht);
+
+	size_t n_objects = hash_get_n_entries(ht);
+	char **objects = calloc(n_objects, sizeof(char *));
```
cgwalters, line 234: Missing `NULL` check

tools/composefs-info.c
```diff
@@ -0,0 +1,294 @@
...
+	fprintf(stderr, "usage: %s [ls|objects|dump] IMAGE\n", argv0);
+}
+
+int main(int argc, char **argv)
+{
+	const char *bin = argv[0];
+	int fd;
+	cleanup_node struct lcfs_node_s *root = NULL;
+	const char *image_path = NULL;
+	const char *command;
```
cgwalters, line 257: So not really related to this exactly but...can I arm twist you into using declare-and-initialize style in this codebase?  I find it much more readable than "big dump of variables at the top of the function".  It plays more nicely with `__attribute__(cleanup)` too because there's no footgun around forgetting to init to `NULL` etc.

tools/cfs-fuse.c
```diff
@@ -973,16 +961,72 @@ static void cfs_release(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *f
...
+	uint64_t n_blocks;
+	uint64_t last_oob_block;
+	bool tailpacked;
+	size_t tail_size;
+	uint32_t raw_blkaddr;
+	off_t oob_size;
+	const uint8_t *tail_data;
+	const uint8_t *oob_data;
+	struct iovec iov[2];
+	int i;
```
cgwalters, line 982: (IMO, this is crying out for declare-and-initialize...)

cgwalters (approved):
A lot of code here, I gave this a semi-skim, only found one actual minor bug (that can be fixed in a followup even); the rest is just nits/questions.

tools/cfs-fuse.c
cgwalters: Not to completely belabor this point but...basically all other sane modern languages allow declaring (and initializing) a variable inside a function.

In the case of this particular function, it would probably be *half the size* if done this way.

I know you work on the Linux kernel which isn't yet using this style, and I know even some others like systemd haven't really adopted it but IMO it really just makes things much more readable.  Particularly in this example!

libcomposefs/lcfs-erofs-internal.h
alexlarsson: Yeah, but I'm not super found of asprintf like this, because I always have to look up the length modifier in the manpage to get it right. 

And also, I always worry about the integer format conversion in varargs calls like this.
Like, name_len is size_t, which could be larger than int in some compilers, and the field width must be int. So, do we need to cast? etc.

libcomposefs/lcfs-writer-erofs.c
alexlarsson: So, a traditional whiteout is a chardev(0,0), and having that in the lowerdir will work as a whiteout for both regular and userxattr overlayfs mounts. Unfortunately they don't work in composefs, because the outer overlayfs eats them.

So, we have a new type of whiteout. If we set an escaped whiteout xattr like `trusted.overlay.overlay.whiteout` on a file in the erofs, then in the overlayfs mount it will show up as `trusted.overlay.whiteout`, and a (new) regular overlayfs mount will treat it as a whiteout. However, to a userxattr overlayfs mount it will not be special in any way, because it looks for `user.overlay.whiteout` files. So, composefs needs to set both of these for the resulting file to always work as a whiteout.

In terms of redundancy. Its not ideal, but it is what we need, and I don't think we'll have tons of whiteouts in practice.

tools/composefs-info.c
alexlarsson: ooops :)

tools/composefs-info.c
alexlarsson: sure

tools/cfs-fuse.c
alexlarsson: Actually, the kernel is moving toward using __attribute__(cleanup), and as a result will be moving towards this style too (at least partially)

alexlarsson: https://lwn.net/ml/linux-kernel/CAHk-=wi-RyoUhbChiVaJZoZXheAwnJ7OO=Gxe85BkPAd93TwDA@mail.gmail.com/

PR #187, opened 2023-09-12
Inline small files

alexlarsson:
Some files are smaller than the size needed for the redirect and the digest, lets allow storing these inline instead.

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -540,6 +540,30 @@ int lcfs_node_set_fsverity_from_fd(struct lcfs_node_s *node, int fd)
 	return lcfs_node_set_fsverity_from_content(node, &_fd, fsverity_read_cb);
 }
 
+static int read_content(int fd, size_t size, uint8_t *buf)
+{
+	int bytes_read;
+
+	while (size > 0) {
+		do
+			bytes_read = read(fd, buf, size);
+		while (bytes_read < 0 && errno == EINTR);
+
+		if (bytes_read == 0)
+			break;
+
+		size -= bytes_read;
+		buf += bytes_read;
+	}
+
+	if (size > 0) {
+		errno = ENODATA;
+		return -1;
+	}
+
+	return 0;
+}
+
 struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
 					     int buildflags)
 {
@@ -548,7 +572,8 @@ struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
 	int r;
 
 	if (buildflags & ~(LCFS_BUILD_SKIP_XATTRS | LCFS_BUILD_USE_EPOCH |
-			   LCFS_BUILD_SKIP_DEVICES | LCFS_BUILD_COMPUTE_DIGEST)) {
+			   LCFS_BUILD_SKIP_DEVICES | LCFS_BUILD_COMPUTE_DIGEST |
+			   LCFS_BUILD_NO_INLINE)) {
 		errno = EINVAL;
 		return NULL;
 	}
@@ -568,17 +593,42 @@ struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
 	ret->inode.st_size = sb.st_size;
 
 	if ((sb.st_mode & S_IFMT) == S_IFREG) {
-		if (sb.st_size != 0 && (buildflags & LCFS_BUILD_COMPUTE_DIGEST) != 0) {
-			int fd = openat(dirfd, fname, O_RDONLY | O_CLOEXEC);
+		bool compute_digest = (buildflags & LCFS_BUILD_COMPUTE_DIGEST) != 0;
+		bool no_inline = (buildflags & LCFS_BUILD_NO_INLINE) != 0;
+		bool is_zerosized = sb.st_size == 0;
+		bool do_digest = !is_zerosized && compute_digest;
+		bool do_inline = !is_zerosized && !no_inline &&
+				 sb.st_size <= LCFS_BUILD_INLINE_FILE_SIZE_LIMIT;
+
+		if (do_digest || do_inline) {
+			cleanup_fd int fd =
+				openat(dirfd, fname, O_RDONLY | O_CLOEXEC);
 			if (fd < 0) {
 				lcfs_node_unref(ret);
 				return NULL;
 			}
-			r = lcfs_node_set_fsverity_from_fd(ret, fd);
-			close(fd);
-			if (r < 0) {
-				lcfs_node_unref(ret);
-				return NULL;
+			if (do_digest) {
+				r = lcfs_node_set_fsverity_from_fd(ret, fd);
+				if (r < 0) {
+					lcfs_node_unref(ret);
+					return NULL;
+				}
+				/* In case we re-read below */
+				lseek(fd, 0, SEEK_SET);
+			}
+			if (do_inline) {
+				uint8_t buf[LCFS_BUILD_INLINE_FILE_SIZE_LIMIT];
+
+				r = read_content(fd, sb.st_size, buf);
+				if (r < 0) {
+					lcfs_node_unref(ret);
+					return NULL;
+				}
+				r = lcfs_node_set_content(ret, buf, sb.st_size);
+				if (r < 0) {
+					lcfs_node_unref(ret);
+					return NULL;
+				}
 			}
 		}
 	}
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -627,6 +627,31 @@ void lcfs_node_set_fsverity_digest(struct lcfs_node_s *node,
 	memcpy(node->digest, digest, LCFS_DIGEST_SIZE);
 }
 
+int lcfs_node_set_content(struct lcfs_node_s *node, const uint8_t *data,
+			  size_t data_size)
+{
+	uint8_t *dup = NULL;
+
+	if (data && data_size != 0) {
+		dup = malloc(data_size);
+		if (dup == NULL) {
+			errno = ENOMEM;
+			return -1;
+		}
+		memcpy(dup, data, data_size);
+	}
+	free(node->content);
+	node->content = dup;
+	node->inode.st_size = data_size;
+
+	return 0;
+}
+
+const uint8_t *lcfs_node_get_content(struct lcfs_node_s *node)
+{
+	return node->content;
+}
+
 const char *lcfs_node_get_name(struct lcfs_node_s *node)
 {
 	return node->name;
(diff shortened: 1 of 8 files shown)


cgwalters:
Quick q, not a full review yet: How does this not change the digest?

alexlarsson:
@cgwalters It changes the digest, but only if you enable it. The default change in mkcomposefs does change the default, but I was thinking we don't have a stable release yet, so it's fine to change at this point.

It doesn't for example change the digest of an ostree generated image, because it doesn't call lcfs_set_content().

libcomposefs/lcfs-writer.c
```diff
@@ -568,17 +592,41 @@ struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
 	ret->inode.st_size = sb.st_size;
 
 	if ((sb.st_mode & S_IFMT) == S_IFREG) {
-		if (sb.st_size != 0 && (buildflags & LCFS_BUILD_COMPUTE_DIGEST) != 0) {
-			int fd = openat(dirfd, fname, O_RDONLY | O_CLOEXEC);
+		bool compute_digest = (buildflags & LCFS_BUILD_COMPUTE_DIGEST) != 0;
+		bool inline_small = (buildflags & LCFS_BUILD_INLINE_SMALL) != 0;
+		bool do_digest = sb.st_size != 0 && compute_digest;
+		bool do_inline = sb.st_size > 0 && sb.st_size <= 64 && inline_small;
```
cgwalters, line 598: Let's hoist this `64` to a `static const` or a `#define` or something that's documented with at least a brief rationale.

libcomposefs/lcfs-writer.c
```diff
@@ -568,17 +592,41 @@ struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
...
-				return NULL;
+			if (do_digest) {
+				r = lcfs_node_set_fsverity_from_fd(ret, fd);
+				if (r < 0) {
+					lcfs_node_unref(ret);
+					return NULL;
+				}
+			}
+			if (do_inline) {
+				uint8_t buf[64];
```
cgwalters, line 615: And then use the same constant here.

libcomposefs/lcfs-writer.c
cgwalters, line 598: It looks odd to me to use `sb.st_size != 0` on one line and then `sb.st_size > 0` in the next, when they're equivalent...

Maybe
```
bool is_zerosized = sb.st_size == 0;
bool do_digest = !is_zerosized && compute_digest;
bool do_inline = !is_zerosized && inline_small && sb.st_size <= LCFS_INLINE_SIZE;
```

or so?

libcomposefs/lcfs-writer.c
```diff
@@ -568,17 +592,41 @@ struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
...
+
+				if (do_digest)
+					lseek(fd, 0, SEEK_SET);
```
cgwalters, line 618: (Unrelated, maybe `lcfs_node_set_fsverity_from_fd` should be changed to maintain the file descriptor position instead?)

cgwalters (commented):
Does mkfs.erofs do this inlining by default?  It looks like it?  Does it also use 64 bytes by default?

It feels like this is exactly the kind of thing we should worry about trying to change "post 1.0" because what if we find out that in some very important workloads it really helps to inline 128 bytes or something?

Or I guess we should aim to match mkfs.erofs, so if it has it off by default we should too...

But if we have these sorts of tunables I fear/suspect we're going to end up needing to compute *multiple* of them and hence allow making it a *dynamic* choice on the client which tunables to set like this.

hsiangkao:
> Does mkfs.erofs do this inlining by default? It looks like it? Does it also use 64 bytes by default?

Yes, `mkfs.erofs` inlines regular files if possible unless passing in -Enoinline_data.  EROFS uses 64-byte inodes just because most people need some timestamp (if a fixed timestamp is passed in like "-T0", mkfs.erofs will use 32-byte inodes instead), but 32-byte doesn't have a way to store ns timestamp (Squashfs keeps in 32-byte inodes but it's just a 4-byte timestamp without nanosecond, and it doesn't support pre-1970 time, so a bit loss on this part. But I'm not sure people really care.)

libcomposefs/lcfs-writer.c
alexlarsson: yeah

libcomposefs/lcfs-writer.c
alexlarsson: I dunno, that would then break use of lcfs_node_set_fsverity_from_fd() with a non-seekable fd (like a pipe).

libcomposefs/lcfs-writer.c
alexlarsson: Yeah, that may be easier to read.

alexlarsson:
> Does mkfs.erofs do this inlining by default? It looks like it? Does it also use 64 bytes by default?

I don't understand this question?  mkfs.erofs will always put all file data in the filesystem image. Where else would it store it? (Although for the erofs file data, there is a question of whether to store it next to the inode, or in a separate block, but that is a different question from this PR.)

For mkcomposefs the question is whether to store the data inline in the erofs image, or whether to use a separate object file in the base directory.
 
Basically, suppose you have an rootfs dir with a 4 byte file, and you run mkcomposefs on it. Do we produce a erofs image that has a 4 byte file? Or do we create a erofs image that has a 0 byte file with an 65 byte overlay.redirect xattr and a 36 byte overlay.metacopy xattr, plus a 4 byte external object file?

> It feels like this is exactly the kind of thing we should worry about trying to change "post 1.0" because what if we find out that in some very important workloads it really helps to inline 128 bytes or something?

There is nothing that fundamentally disallows an inlining limit of 128 bytes with the proposed API. The only place that currently hardcodes the 64 byte limit is the "convert this real directory to a lcfs-node" helper. Ostree for example doesn't use that, and could pick whatever inline limit it wanteds (if it wanted too). And, we could easily add an option to lcfs_load_node_from_file() to use a different limit if we wanted. It would be up to each user of the tools/APIs whether they wanted to enable said flag, with the result that the generated images would get a different digest.

That said, I don't think the limit is all that important once you reach some slightly larger size, because the overhead starts to be less ridiculous.  So, I don't think the precise limit is all that important, as long as e.g. files smaller than a sha256 digest are inlined.

> Or I guess we should aim to match mkfs.erofs, so if it has it off by default we should too...

mkfs.erofs never ever puts any files in the basedir though. The concept doesn't even exist.

> But if we have these sorts of tunables I fear/suspect we're going to end up needing to compute _multiple_ of them and hence allow making it a _dynamic_ choice on the client which tunables to set like this.

I agree that we should limit tunables, which is why I made this the default for mkcomposefs. Maybe we should drop the "--no-inline" option though?

giuseppe (approved):
LGTM

cgwalters:
> (Although for the erofs file data, there is a question of whether to store it next to the inode, or in a separate block, but that is a different question from this PR.)

Yes, I was referring to the inode.

> For mkcomposefs the question is whether to store the data inline in the erofs image, or whether to use a separate object file in the base directory.

Right but I think you answered that with this:

> So, I don't think the precise limit is all that important, as long as e.g. files smaller than a sha256 digest are inlined.

Yes, this makes sense to me - so I'd argue: let's do it by default, take another "checksum epoch" and try to commit to your suggested 64 bytes by default.

cgwalters:
Just going to link this one for you and me to have a blast from the past: https://ometer.com/preferences.html - it's an argument not to add an option.

alexlarsson:
I'm gonna merge this one and do the cfs-fuse change as part of !185 which has some other fuse stuff.

alexlarsson:
> > (Although for the erofs file data, there is a question of whether to store it next to the inode, or in a separate block, but that is a different question from this PR.)
> 
> Yes, I was referring to the inode.
> 
> > For mkcomposefs the question is whether to store the data inline in the erofs image, or whether to use a separate object file in the base directory.
> 
> Right but I think you answered that with this:
> 
> > So, I don't think the precise limit is all that important, as long as e.g. files smaller than a sha256 digest are inlined.
> 
> Yes, this makes sense to me - so I'd argue: let's do it by default, take another "checksum epoch" and try to commit to your suggested 64 bytes by default.

Well, we sort of already have a checksum epoch for the support of xattr escapes. But, I think we're now essentially ready to have a real frozen format.

libcomposefs/lcfs-writer.c
cgwalters: Simpler: why not move this `lseek` underneath the previous `do_digest` conditional?

(I know this is a minor nit, I just wondered how this lseek was related to digests)

tests/dumpdir
```diff
@@ -0,0 +1,47 @@
...
+    rel = os.path.relpath(file, root)
+    s = os.lstat(file)
+    nlink = s.st_nlink;
+    if args.no_nlink:
+        nlink = 1
+    print(f"{shlex.quote(rel)} {s.st_mode} {nlink} {s.st_uid}:{s.st_gid} {s.st_rdev} {s.st_mtime_ns}",end="")
+    if stat.S_ISREG(s.st_mode):
+        digest = hashlib.sha256(open(file,'rb').read()).hexdigest()
+        print(f" {s.st_size} sha256:{digest}",end="")
+    if stat.S_ISLNK(s.st_mode):
```
cgwalters, line 23: nit: `elif`

libcomposefs/lcfs-writer.c
```diff
@@ -540,6 +540,29 @@ int lcfs_node_set_fsverity_from_fd(struct lcfs_node_s *node, int fd)
...
+
+	while (size > 0) {
+		do
+			bytes_read = read(fd, buf, size);
+		while (bytes_read < 0 && errno == EINTR);
+
+		if (bytes_read == 0)
+			break;
+
+		size -= bytes_read;
```
cgwalters, line 555: Aren't we also missing `buf += size` or so?

alexlarsson: yeah

libcomposefs/lcfs-writer.c
alexlarsson: Well, then we would seek always after set_fsverity, even if we were not gonna inline it... 
Still, maybe less weird code-wise. Lemme do that.

tools/mkcomposefs.c
```diff
@@ -354,7 +355,7 @@ static int fill_payload(struct lcfs_node_s *node, const char *path, size_t len,
 static void usage(const char *argv0)
 {
 	fprintf(stderr,
-		"usage: %s [--use-epoch] [--skip-xattrs] [--absolute] [--by-digest] [--digest-store=path] [--print-digest] [--print-digest-only] [--skip-devices] [--compute-digest] SOURCEDIR IMAGE\n",
+		"usage: %s [--use-epoch] [--skip-xattrs] [--absolute] [--no-inline] [--by-digest] [--digest-store=path] [--print-digest] [--print-digest-only] [--skip-devices] [--compute-digest] SOURCEDIR IMAGE\n",
```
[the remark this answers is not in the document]
alexlarsson: dropping it for now

PR #190, opened 2023-09-15
composefs-images: Add missing-object command and support multiple images

alexlarsson:
If you pass in --basedir=objectdir and N images, then the `missing-object` commands will print all the object in any of the images that are don't exist in the basedir.


giuseppe (approved):
LGTM

PR #191, opened 2023-09-18
More tests

alexlarsson:
This adds some fusa/tools tweaks and then some initial work on randomized image testing.
For now it only uses fuse so we can run this during unit tests, but I want to follow up with similar integration tests that do an actual kernel mount.

### DIFF
--- a/libcomposefs/lcfs-writer-erofs.c
+++ b/libcomposefs/lcfs-writer-erofs.c
@@ -881,7 +881,7 @@ static int write_erofs_inode_data(struct lcfs_ctx_s *ctx, struct lcfs_node_s *no
 		i.i_uid = lcfs_u32_to_file(node->inode.st_uid);
 		i.i_gid = lcfs_u32_to_file(node->inode.st_gid);
 		i.i_mtime = lcfs_u64_to_file(node->inode.st_mtim_sec);
-		i.i_mtime_nsec = lcfs_u64_to_file(node->inode.st_mtim_nsec);
+		i.i_mtime_nsec = lcfs_u32_to_file(node->inode.st_mtim_nsec);
 
 		if (type == S_IFDIR) {
 			if (node->erofs_n_blocks > 0) {
--- a/tests/gendir
+++ b/tests/gendir
@@ -1,0 +1,195 @@
+#!/usr/bin/python3
+
+import argparse
+import hashlib
+import os
+import random
+import shlex
+import shutil
+import stat
+import string
+import sys
+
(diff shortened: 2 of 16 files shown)


alexlarsson:
Hmm, seems like fsck.erofs is not big endian safe, at least not the version in ubuntu.

libcomposefs/lcfs-writer-erofs.c
```diff
@@ -881,7 +881,7 @@ static int write_erofs_inode_data(struct lcfs_ctx_s *ctx, struct lcfs_node_s *no
 		i.i_uid = lcfs_u32_to_file(node->inode.st_uid);
 		i.i_gid = lcfs_u32_to_file(node->inode.st_gid);
 		i.i_mtime = lcfs_u64_to_file(node->inode.st_mtim_sec);
-		i.i_mtime_nsec = lcfs_u64_to_file(node->inode.st_mtim_nsec);
+		i.i_mtime_nsec = lcfs_u32_to_file(node->inode.st_mtim_nsec);
```
cgwalters, line 884: Ah yes, good old C automatic integer promotion rules, here to blow off our feet.

So I tested this, and it finds the problem:

```
diff --git a/libcomposefs/lcfs-internal.h b/libcomposefs/lcfs-internal.h
index f9cab47..749dbe6 100644
--- a/libcomposefs/lcfs-internal.h
+++ b/libcomposefs/lcfs-internal.h
@@ -70,10 +70,10 @@ static inline uint32_t lcfs_u32_to_file(uint32_t val)
 	return htole32(val);
 }
 
-static inline uint64_t lcfs_u64_to_file(uint64_t val)
-{
-	return htole64(val);
-}
+#define lcfs_u64_to_file(v) ({ \
+	_Static_assert(sizeof(v) == sizeof(uint64_t)); \
+	htole64(v); \
+})
 
 static inline uint16_t lcfs_u16_from_file(uint16_t val)
 {
```

Opinions?

tests/gendir
```diff
@@ -0,0 +1,195 @@
+#!/usr/bin/python3
```
cgwalters, line 1: This repository is now up to 3 distinct languages  (C, shell script, and Python).  Personally, I would vote that eventually of using Python and shell script for tests we use Rust.  I've had decent success with this, see e.g. https://github.com/ostreedev/ostree/blob/878d601665fd15781adf774d99dd86e9a172cb92/tests/inst/src/destructive.rs#L477 shows how Rust's macros can let one execute shell commands ([xshell crate](https://docs.rs/xshell/latest/xshell/)) with near-equal ergonomics to bash, but more safety than Python.

cgwalters (approved):
Overall looks sane, but I didn't deep dive into the tests or the xattr changes.

libcomposefs/lcfs-writer-erofs.c
alexlarsson: This may hit problematic cases where we e.g. pass an int to u64_to_file(), but I'm fine with having those explicitly add a cast as needed.

tests/gendir
alexlarsson: I love rust for production code, but I'm not sure how well it would fit with code that is a bit more throwaway and like tests. Seems like it may raise the bar for people to add new testcases. I'd be willing to be convinced though.

libcomposefs/lcfs-writer-erofs.c
alexlarsson: https://github.com/containers/composefs/pull/192

hsiangkao:
> Hmm, seems like fsck.erofs is not big endian safe, at least not the version in ubuntu.

I don't quite have an environment to test big-endian... I know I could setup a QEMU environment, but honestly currently I don't have enough time on this now..

PR #192, opened 2023-09-19
Add static asserts that endian conversions are of right size

alexlarsson:
This makes it easy to detect accidental cases where you get this wrong, and make it more explicit where we're casting to a different size.

### DIFF
--- a/libcomposefs/lcfs-writer-erofs.c
+++ b/libcomposefs/lcfs-writer-erofs.c
@@ -810,6 +810,8 @@ static int write_erofs_inode_data(struct lcfs_ctx_s *ctx, struct lcfs_node_s *no
 	compute_erofs_xattr_counts(node, &n_shared_xattrs, &unshared_xattrs_size);
 	xattr_size = xattr_erofs_inode_size(n_shared_xattrs, unshared_xattrs_size);
 	xattr_icount = xattr_erofs_icount(xattr_size);
+	if (xattr_icount > UINT16_MAX)
+		return -EINVAL;
 
 	version = node->erofs_compact ? 0 : 1;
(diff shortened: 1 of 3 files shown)


libcomposefs/lcfs-writer-erofs.c
```diff
@@ -838,18 +840,19 @@ static int write_erofs_inode_data(struct lcfs_ctx_s *ctx, struct lcfs_node_s *no
 	if (node->erofs_compact) {
 		struct erofs_inode_compact i = { 0 };
 		i.i_format = lcfs_u16_to_file(format);
-		i.i_xattr_icount = lcfs_u16_to_file(xattr_icount);
-		i.i_mode = lcfs_u16_to_file(node->inode.st_mode);
-		i.i_nlink = lcfs_u16_to_file(node->inode.st_nlink);
-		i.i_size = lcfs_u32_to_file(size);
+		i.i_xattr_icount = lcfs_u16_to_file((uint16_t)xattr_icount);
+		i.i_mode = lcfs_u16_to_file((uint16_t)node->inode.st_mode);
+		i.i_nlink = lcfs_u16_to_file((uint16_t)node->inode.st_nlink);
```
cgwalters, line 845: For cases like this, maybe we should have a helper equivalent to [this Rust API](https://doc.rust-lang.org/std/convert/trait.TryFrom.html#impl-TryFrom%3Cu32%3E-for-i16)?

cgwalters (approved):

libcomposefs/lcfs-writer-erofs.c
alexlarsson: In this particular case we already checked in lcfs_fits_in_erofs_compact() that it should fit, so maybe an version that asserts it fits is better. However, in other cases we could have an error reporting case.

PR #195, opened 2023-09-22
Simplify mkcomposefs

alexlarsson:
This moves some code from mkcomposefs to lcfs_build() to:
 * Create by-digest payloads
 * Create symlink payloads
And then uses it

It also simplifies the mkcomposefs arguments to always compute digest and by-digest payloads, dropping the --absolute option and related things, as we couldn't think of a usecase for it right now. If we have a more precise usecase later we can add it back in a way that makes sense for the usecase.


giuseppe (approved):
LGTM

PR #196, opened 2023-09-22
Add more docs

alexlarsson:
This adds the simplify commits that was missing from https://github.com/containers/composefs/pull/195, and updates the docs for mkcomposefs and mount.composefs.


giuseppe (approved):
LGTM

tools/mountcomposefs.c
```diff
@@ -229,14 +238,8 @@ int main(int argc, char **argv)
 
 	options.expected_fsverity_digest = opt_digest;
 
-	if (opt_verity && opt_noverity) {
-		printexit("Incompatible options verity, noverity\n");
-	}
-
-	if (opt_verity)
+	if (opt_verity || opt_digest != NULL)
```
cgwalters, line 241: This is one of the bigger changes here right?  I think this makes sense...no point in verifying the digest without also doing underlying verity I guess?

cgwalters (approved):

tools/mountcomposefs.c
alexlarsson: Yeah, i think it would be surprising if you successfully mounted against a particular digest, but then you can still read back modified data from the mount.

PR #199, opened 2023-09-22
Update for release 0.9.0

alexlarsson:
Bump version to 0.9.0
    
    This is the first stable release of composefs. Starting now, we
    guarantee a stable library ABI and a binary stable file format. The
    later means that any image build from an identical lcfs_node tree and
    identical write options, will produce a file that is binary identical
    to a later run even with a different version. The same is true for
    an mkcomposefs run with the same options.
    
    Major changes since 0.1.4:
    
    * All required overlayfs xattr changes are now upstream and the
      corresponding image generation changes have been made in
      composefs. This includes support for escaping overlayfs xattrs and
      whiteouts for nested overlayfs mounts.
    
    * fs-verity built-in signature support was dropped in favour of
      userspace signatures.
    
    * The erofs iamges now uses the new bloom filter for faster xattr
       lookups. This is backward compatible and old erofs version still
       work (sans the speedup).
    
    * Files can now be inlined in the erofs image to avoid overhead of
      using redirections for small files.
    
    * There is a new API to regenerate a lcfs_node tree from a composefs
      image file.
    
    * There is a new composefs-info tool that lets you dump info about
      images, including what objects it refers to and which ones are
      missing from a given basedir.
    
    * Various fixes, cleanups and new tests

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,14 +1,27 @@
 SUBDIRS=libcomposefs tools tests
 
+CLEANFILES=
+
+MANPAGES=\
+	man/mount.composefs.md \
+	man/mkcomposefs.md
+
 EXTRA_DIST=\
 	composefs.pc.in \
 	composefs.spec.in \
-	composefs.spec
+	composefs.spec \
+	${MANPAGES}
 
 pkgconfig_DATA = composefs.pc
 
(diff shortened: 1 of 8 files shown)


giuseppe (approved):
LGTM

Issue #200, opened 2023-09-23
portability issues found while packaging


Kiskae:
While preparing packaging for 0.9.0 over at https://github.com/NixOS/nixpkgs/pull/256892 I ran into a few potential portability issues while trying various cross-compilations and thought to report them.

- `#import <endian.h>` in `libcomposefs/lcfs-fsverity.c` - [not supported by darwin or freebsd](https://stackoverflow.com/a/58418801/1452094)
- `#import <error.h>` in `tools/*` - [gnu extension, not available when using musl](https://wiki.gentoo.org/wiki/Musl_porting_notes#error.h:_No_such_file_or_directory)
- `os.setxattr` in `tests/gendir` - throws an exception if the file system the builder is running on does not support xattrs:
```
composefs> Traceback (most recent call last):
composefs>   File "/build/source/tests/./gendir", line 193, in <module>
composefs>     make_dir(root, dirs)
composefs>   File "/build/source/tests/./gendir", line 163, in make_dir
composefs>     set_user_xattr(path)
composefs>   File "/build/source/tests/./gendir", line 104, in set_user_xattr
composefs>     os.setxattr(path, name, value, follow_symlinks=False)
composefs> OSError: [Errno 95] Operation not supported: b'/build/lcfs-test.pTv96t/root'
```

While the file system itself only works on linux, it would be nice if the tooling to create or inspect a composefs image works on other unix variants.

Kiskae:
Not a portability issue, but `descriptor.data_size_be = htole64(ctx->file_size);` indicates that should probably be `data_size_le` to avoid confusion.

alexlarsson:
Lemme have a quick look at this before finalizing the 0.9.0 release

alexlarsson:
@Kiskae Btw, is there any interest in using composefs for the nix store? Seems like it could make a good match?

Kiskae:
I'm not involved with the nix implementation so I wouldnt be able to tell you if there is interest on that side.

Regarding the use of composefs for the nix store I'm not sure if it would work for the global store since it changes fairly frequently (every time someone realizes a derivation), but there might be a use in the sandboxed builders to create a local version of the nix store without exposing all the build dependencies on the global store.

alexlarsson:
I meant more for the local store to do e.g. deduplication, similar to the hardlink optimization that nix does, but more powerful.

Kiskae:
With global store I meant the nix store at /nix/store, which I presume you meant with local store.

[...]

PR #202, opened 2023-09-25
Fix various portability issues

alexlarsson:
This should fix the issues pointed out in #200 (although I never tried to actually build on any other os).


Kiskae:
After [removing `AC_FUNC_ALLOC/AC_FUNC_REALLOC` from autoconf](https://stackoverflow.com/questions/67116092/ld-problem-rpl-malloc-cross-compiling-arm) I was able to compile a static binary with these changes.

> ./result/bin/mkcomposefs: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, not stripped

Those autoconf statements redirect the code through a `rpl_malloc` function, which does to not exist in musl/some clang stdlibs.

Cannot confirm whether the darwin compilation builds at the moment.

alexlarsson:
rpl_malloc is meant to be supplied manually by the code if, but we don't do this atm. I don't think we depend on the gnu malloc(0) behaviour tho, so lets just drop these.

Kiskae:
- `test-random-fuse` works on filesystems without xattrs
- static linking with musl works
- compilation on darwin does not work
  - it looks like apple simply does not supply `endian.h`, you need to provide your own wrapper to `libkern/OSByteOrder.h`
- compilation on clang fails with `cfs-fuse.c:57:19: error: format string is not a string literal `
  - disabling fuse simply moves the error to `mountcomposefs.c:46:19: error: format string is not a string literal`

alexlarsson:
>     * compilation on darwin does not work
>      
>       * it looks like apple simply does not supply `endian.h`, you need to provide your own wrapper to `libkern/OSByteOrder.h`

This will have to be done by someone with darwin then.
 
>     * compilation on clang fails with `cfs-fuse.c:57:19: error: format string is not a string literal `
>       
>       * disabling fuse simply moves the error to `mountcomposefs.c:46:19: error: format string is not a string literal`

Fixed

Kiskae (approved):
Can confirm that both builds with musl or clang compile successfully.

Although it might be an idea to add this to `libcomposefs/Makefile-lib.am` so the current version gets set as SONAME:
```Make
libcomposefs_la_LDFLAGS = -release @VERSION@
```

alexlarsson:
> Can confirm that both builds with musl or clang compile successfully.
> 
> Although it might be an idea to add this to `libcomposefs/Makefile-lib.am` so the current version gets set as SONAME:
> 
> ```makefile
> libcomposefs_la_LDFLAGS = -release @VERSION@
> ```

The version number of the release is more of a human consumable moniker. We don't want to tread it as an actual soname. For example, we want to be able to go to 1.0.0 without breaking the library ABI.

We should probably have *some* minor soname changes tho..

tools/composefs-from-json.c
```diff
@@ -536,20 +541,20 @@ static int fill_file(const char *typ, struct lcfs_node_s *root,
 		cleanup_free uint8_t *buf = malloc(buf_size);
 
 		if (buf == NULL) {
-			error(0, 0, "malloc");
+			fprintf(stderr, "malloc");
```
cgwalters, line 544: This is missing a newline though.  Why are we printing here?  Supporting custom strings really requires going to a whole GError/Result style error handling; in baseline C we can only set errno right?

cgwalters (changes requested):

tools/composefs-from-json.c
alexlarsson: Well, its a tool, not a library, so while not ideal, printing here isn't completely wrong. It should probably exit on printing tho.

alexlarsson: I added the newline, and we do generally exit after the print, the tool just prints two error lines.

I think this is fine for now, because composefs-from-json really is an internal thing at this point.

giuseppe (approved):
LGTM

PR #203, opened 2023-09-26
Add soname for libcomposefs

alexlarsson:
This starts at 1.0.0 to signal incompatibilty with all previous (unstable) releases. At this point we strive to be ABI compat, and update the soname micro for non-api changes, and the minor for api additions.


giuseppe (approved):
LGTM

PR #211, opened 2023-10-06
WIP: tools: Add an `fsck-objects` option

doanac:
This is a total shot in the dark. I was playing around with corrupting object content and how to detect it. Thought the idea of an `fsck` type tool might be useful to the project. If it is, I'm happy to clean things up as you see fit. Otherwise, feel free to close this.

### DIFF
--- a/tools/composefs-info.c
+++ b/tools/composefs-info.c
@@ -226,7 +226,7 @@ static const char *abs_to_rel_path(const char *path)
 	return path;
 }
 
-static void get_objects(struct lcfs_node_s *node, PrintData *data, int basedir_fd)
+static void get_objects(struct lcfs_node_s *node, PrintData *data, int basedir_fd, bool fsck)
 {
 	uint32_t mode = lcfs_node_get_mode(node);
 	uint32_t type = mode & S_IFMT;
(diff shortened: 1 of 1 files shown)


alexlarsson:
Yeah, that does seems useful. Just iterating over an object store and ensure that all filenames properly checksum to their names is a first step. Would also be nice if it could re-enable fs-verity on the files (if possible), which is useful when you copy basedirs between machines.

alexlarsson:
I'm not sure this is the right way this should work though. Generally what you have is a lot of composefs images, and a shared directory of backing files. We want a tool that takes the path of the backing directory and enumerates all the files in it, verifying that their content matches the name.

So you'd have `composefs-fsck /path/to/dir`.

doanac:
> I'm not sure this is the right way this should work though. Generally what you have is a lot of composefs images, and a shared directory of backing files. We want a tool that takes the path of the backing directory and enumerates all the files in it, verifying that their content matches the name.

I can see value in that also. However, In the tooling I've been toying around with, I wanted to understand which compose-fs image was affected. I guess I could still do a global fsck, then dump objects for each image and find that.

If y'all are open to a full-blown "composefs-fsck" command (i have a feeling shoving this in `composefs-info` might not be very clean), I'd be happy to try that approach. I also know you folks are working really hard right now and might not want to coach someone through a feature like this. Feel free to say "not right now" or just do it yourself when the time is right for you.

This is a great project - I think there are going to be a lot of interesting things that come out of it.

alexlarsson:
This is a bit stale, i'm closing it for now, but would still be nice to have the extended tool.

PR #215, opened 2023-10-10
Add mkcomposefs --from-file

alexlarsson:
This supports creating a composefs image from a simple line-per-file format. 

The format used is the same as the one that `composefs-info dump` generates, which allows conversion to from textual format. It is also a format that is very easy to produce and e.g. stream to mkcomposefs from other languages.  

Even though very simple to generate and parse, the format is quite compact. For example, the f36-x86_64-silverblue.json.gz file is 33M uncompressed and generates a 19M composefs image. But the dump file for it is just 16M.


alexlarsson:
@giuseppe I think containers/storage could generate this and stream it into `mkcomposefs --from-file -`.

alexlarsson:
Added some documentation too..

giuseppe (approved):
LGTM

Issue #213, opened 2023-10-07
mount.composefs fails with `Failed to mount composefs example.cfs: No such file or directory` when upperdir has a comma in the name


eriksjolund:
This issue is probably caused by a Linux kernel regression.

**Steps to reproduce the issue:**

1. `sudo -i`
2. `mkdir basedir`
3. `mkdir workdir`
4. `mkdir upper,dir`
5. `mkdir mnt`
6. `mkcomposefs ./basedir example.cfs`
7.  Run 
    ```
    # mount.composefs -t composefs \
       -o 'basedir=./basedir,workdir=./workdir,upperdir=./upper\,dir' example.cfs ./mnt
    mount.composefs: Failed to mount composefs example.cfs: No such file or directory
    #
    ```

**Describe the results you received:**

Command in step 7 fails.

**Describe the results you expected:**

I had expected the command in step 7 to succeed.

**About the system**

```
# uname -r
6.6.0-0.rc2.20230919git2cf0f7156238.21.fc40.aarch64
# rpm-ostree status
State: idle
Deployments:
● fedora:fedora/aarch64/coreos/rawhide
                  Version: 40.20230921.91.0 (2023-09-21T14:07:56Z)
                   Commit: d8eab688f9726a1aac5d55922a0d205c03fbf243d30d7e9c4e280f0190a2abe0
             GPGSignature: Valid signature by 115DF9AEF857853EE8445D0A0727707EA15B79CC
# 
```

**Additional note 1**

This issue is probably caused by a Linux kernel regression

Quote: _"Up to and including kernel 6.4.15, it was possible to have commas in 
the lowerdir/upperdir/workdir paths used by overlayfs, provided they were 
escaped with backslashes:"_

See
https://lore.kernel.org/all/8da307fb-9318-cf78-8a27-ba5c5a0aef6d@alum.mit.edu/

**Additional note 2**

I could not  reproduce the bug on an older Fedora CoreOS version. 
About the older system:
```
[core@localhost ~]$ rpm-ostree status
State: idle
AutomaticUpdatesDriver: Zincati
  DriverState: inactive
Deployments:
● fedora:fedora/aarch64/coreos/next
                  Version: 38.20230310.1.0 (2023-03-10T22:51:50Z)
                   Commit: b0fdf736cdbbd3971380d5549635e30155f07af6100925d987de623b4722637f
             GPGSignature: Valid signature by 6A51BBABBA3D5467B6171221809A8D7CEB10B464
[...]

PR #218, opened 2023-10-11
Switch to new mount api for ovl on kernel 6.5 to fix escaping

alexlarsson:
This is required on kernel >= 6.5 to handle escaped commas in upper or lowerdirs due to an overlayfs change in escape handling behaviour  caused by the switch to the new mount api internally.
    
For anything post 6.5 we fallback to the old legacy implementation which supports escaping.

This fixes https://github.com/containers/composefs/issues/213

### DIFF
--- a/libcomposefs/lcfs-mount.c
+++ b/libcomposefs/lcfs-mount.c
@@ -334,6 +334,75 @@ static char *compute_lower(const char *imagemount,
 	return lower;
 }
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
I tested this under strace with 6.5.6-200.fc38.x86_64 (successfully used new api, worked with comma in upperdir), and 6.4.15-200.fc38.x86_64 (falls back to new api both when comma in upperdir and without, both succeeding).

giuseppe (approved):
LGTM

PR #220, opened 2023-10-12
composefs-from-json; Include readv/writev in seccomp list

### LINKED ISSUE #206: composefs-from-json: seccomp breaks stdio on musl
On musl stdio is implemented using `readv/writev`: https://wiki.musl-libc.org/functional-differences-from-glibc.html

Since these syscalls aren't included in the seccomp filter it breaks stdio and since this also breaks stderr it fails to print an error and silently ends the program.

alexlarsson:
Musl uses readv/writev for stdio, so we need to allow them for the tests to pass.

Fixes #206


giuseppe (approved):
LGTM

PR #221, opened 2023-10-12
Use a sandbox with mkcomposefs --from-file

alexlarsson:
This is for similar reasons as it being used with composefs-from-json.


alexlarsson:
==5399==Can't open /proc/5384/task for reading.
==5384==LeakSanitizer has encountered a fatal error.

I wonder if we need to add a /proc to the sandbox. That should be relatively safe.

cgwalters:
I'm not opposed to this but I find it strange...I think most sandboxing like this is much better done externally.  I mean we're talking about having this code be run from a thing that's a container runtime already.

alexlarsson:
> I'm not opposed to this but I find it strange...I think most sandboxing like this is much better done externally. I mean we're talking about having this code be run from a thing that's a container runtime already.

It is somewhat uncommon, but not totally. Its more like the openbsd pledge thing, where you do some setup, including opening all fd:s that you need, and then limit yourself for the actual work.  For example, we run the parser code without any access to the open/creat syscalls, which isn't really possible to do from the caller, as we need to open things in the ssetup.

That said, maybe it does overcomplicate things... For example with the CI failure...

alexlarsson:
closing this for now, not sure its right

PR #225, opened 2023-10-16
Start collecting Fedora review fixes

smooge:
- Take in fixes from reviwers to fix man page compression types
- Take in fixes from reviwers to move licenses to lib subpackage

### DIFF
--- a/composefs.spec.in
+++ b/composefs.spec.in
@@ -53,6 +53,7 @@ rm -rf %{buildroot}%{_libdir}/libcomposefs.la
 %{_libdir}/pkgconfig/%{name}.pc
 
 %files libs
+%license COPYING COPYING.LIB COPYING.LESSERv3 COPYINGv3 LICENSE.Apache-2.0 BSD-2-Clause.txt
 %{_libdir}/libcomposefs.so.*
 
 %files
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Looks sane to me but honestly I feel like keeping spec files upstream has generally just involved duplication and pain (e.g. how one can't use autochangelog in C9S) so it's easier to just keep these things downstream.

smooge:
Understood and mostly agreed. I am not sure if this is 'helping' as I expect that if someone wanted this for SuSE or Mageia or CBL-Mariner they would need to make more changes. I will stop making PR's from other changes found in the review unless the change really fixes something missed before

cgwalters:
OK, want to change this PR to just remove the spec file and hence go to one fewer source of truth?

smooge:
Sorry I missed your message. I am going to talk with @alexlarsson on this as I think he was wanting the spec for CI or something. I will update PR when I figure that out.

smooge:
I tested this MR against f38-qemu-developer-regular.x86_64.json and cs9-qemu-developer-regular.x86_64.json

alexlarsson:
I think its nice to have a good spec file for now, so other distros have a sane start when they start packaging it. I agree that it is sort of a paint to keep it updated though.

I'll just merge this for now, and we can reconsider dropping it at a later date.

PR #226, opened 2023-10-26
Drop composefs-from-json

alexlarsson:
This moves the test assed to the dump format (producing identical images) and drops composefs-from-json. This allows us to drop a bunch of dependencies and close some bugs, and also this is not used anymore by podman.


cgwalters (approved):

nikstur:
This would've been a really cool feature for NixOS because we do not have access to actual users (as in entries in/etc/passwd) in the nix-build sandbox but have all the information in Nix (which is easy to convert to json).

PR #227, opened 2023-11-07
Use go-md2man for building manpages

yselkowitz:
This allows avoiding the pandoc build dependency in downstream distributions (which is important for RHEL which does not provide pandoc) at the price of upstream developers requiring it.

The alternative would be to add the manpages to source control, so that pandoc would only be needed when modifying the markdown sources.

This is needed to remove the unwanted pandoc from the ELN buildroot while still providing man pages.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -28,11 +28,11 @@ if ENABLE_MAN
 
 man/%.1: man/%.md
 	mkdir -p man
-	${PANDOC} $+ -s -t man > $@
+	${MD2MAN} -in $^ -out $@
 
 man/%.5: man/%.md
 	mkdir -p man
-	${PANDOC} $+ -s -t man > $@
+	${MD2MAN} -in $^ -out $@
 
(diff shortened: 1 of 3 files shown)


cgwalters:
Can we use `go-md2man` instead?

yselkowitz:
`golang-github-cpuguy83-md2man` is already a dependency of several other container tools packages, so that would be acceptable.

cgwalters (approved):

yselkowitz:
Downstream backport: https://src.fedoraproject.org/rpms/composefs/pull-request/1

yselkowitz:
One thing that I missed is that ELN (RHEL 10) has disabled golang on i686, so there is no go-md2man there.  While composefs includes a library, I'm not sure what if anything will use that. Does there need to be a composefs build at all on i686, or do we need to simply disable manpages thereon?

cgwalters:
> Does there need to be a composefs build at all on i686

There is no need for composefs on i686.

PR #228, opened 2023-11-10
Update EROFS documentation link

hsiangkao:
Update [the link](https://erofs.docs.kernel.org) which contains more information about the latest status of EROFS although the website is still under construction for now.

Besides, the content of the old `.txt` one has been outdated for years since kernel documentation were switched into the reStructuredText format then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ mountable filesystem trees, stacking on top of an underlying
 The key technologies composefs uses are:
 
 - [overlayfs](https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt) as the kernel interface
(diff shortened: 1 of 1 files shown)


hsiangkao:
Hi @cgwalters @alexlarsson !

Would you mind to give some hints if it is a good idea to update this?
Since I will also write more detailed cases like Composefs later in the new documentation site, thanks!

cgwalters (approved):
Thanks!

cgwalters:
> Would you mind to give some hints if it is a good idea to update this?  Since I will also write more detailed cases like Composefs later in the new documentation site, thanks!

Seems like a good idea to me!

PR #229, opened 2023-11-10
Drop i686 from spec

yselkowitz:
Only certain userspace libraries are still needed for compatibility on i686, and composefs is certainly not among them.

This is needed to fix the ELN build after #227, as golang (and hence go-md2man) are disabled there on i686.

### DIFF
--- a/composefs.spec.in
+++ b/composefs.spec.in
@@ -10,6 +10,9 @@ Source0:        https://github.com/containers/composefs/releases/download/v%{ver
 BuildRequires:  gcc automake libtool openssl-devel go-md2man fuse3-devel
 Requires:       %{name}-libs = %{version}-%{release}
 
+# https://fedoraproject.org/wiki/Changes/EncourageI686LeafRemoval
+ExcludeArch:    %{ix86}
+
 %description
 Tools to handle creating and mounting composefs images. The composefs
 project combines several underlying Linux features to provide a very


cgwalters (approved):
(I'd personally lean towards removing the spec file from upstream entirely now FWIW and just do this stuff downstream precisely because it's downstream stuff that doesn't make sense lifecycled upstream)

yselkowitz:
Upstream spec files are used with the packit workflow, not sure if that's your plan or not.

yselkowitz:
Downstream: https://src.fedoraproject.org/rpms/composefs/pull-request/2

cgwalters:
I'm pretty sure packit can be configured to find things in downstream too

cgwalters:
OK more fallout in:

```
$ sudo dnf repoquery --whatrequires rpm-ostree
cockpit-ostree-1:197-1.fc39.noarch
gnome-software-rpm-ostree-0:45.0-1.fc39.x86_64
greenboot-0:0.15.4-3.fc39.x86_64
osbuild-ostree-0:100-1.fc39.noarch
$
```

I imagine there's no problems with dropping i686 for those either, but it will just take some grunt work.

cgwalters:
xref https://src.fedoraproject.org/rpms/rpm-ostree/pull-request/56

dougsland:
+1 for drop (little late)

PR #230, opened 2023-12-02
libcomposefs/lcfs-writer.c: fix build without reallocarray

ffontaine:
Use `realloc` instead of `reallocarray` to avoid the following build failure with uclibc:

```
composefs/libcomposefs/lcfs-writer.c: In function 'lcfs_node_add_child': composefs/libcomposefs/lcfs-writer.c:886:24: error: implicit declaration of function 'reallocarray' [-Werror=implicit-function-declaration]
  886 |         new_children = reallocarray(parent->children, sizeof(*parent->children),
      |                        ^~~~~~~~~~~~
composefs/libcomposefs/lcfs-writer.c:886:22: error: assignment to 'struct lcfs_node_s **' from 'int' makes pointer from integer without a cast [-Werror=int-conversion]
  886 |         new_children = reallocarray(parent->children, sizeof(*parent->children),
      |                      ^
```

Fixes:
 - http://autobuild.buildroot.org/results/38470213c86091d14c59716cddcccd555f875705


cgwalters (changes requested):
But the API exists for a reason; from `man reallocarray`:

> However, unlike that realloc() call, reallocarray() fails safely in the case where the multiplication would overflow.  If such an overflow occurs, reallocarray() returns an error.

I would offhand bet that there are at least 50 similar patches for projects for uclibc compatibility, and particularly given the security/correctness properties in play here, there seems to be a strong argument that uclibc should gain this API, not that we should drive every project to the lowest common denominator.

(Tangentially of course, stuff like this IMO is a strong argument for Rust where not only are things memory safe by default, there's a *much* more useful truly standard library)

cgwalters:
It looks like systemd carries a reimplementation:

```
#if !HAVE_REALLOCARRAY
_alloc_(2, 3) static inline void *reallocarray(void *p, size_t need, size_t size) {
        if (size_multiply_overflow(size, need))
                return NULL;

        return realloc(p, size * need ?: 1);
}
#endif
```

So I guess we can too.  But we should do that instead of just falling back.

cgwalters:
> By the way, would that mean, that the other 2 calls to realloc in composefs are also unsecured and should be replaced by reallocarray?

I would use the term "insecure" if they can be reached by untrusted input, which may be actually be possible.  Glancing at the two, I think we are missing a fixed capacity on number of xattrs for example.

cgwalters:
PR for the xattr bits in https://github.com/containers/composefs/pull/232 which should ensure we can never overflow there.

libcomposefs/lcfs-writer.c
```diff
@@ -958,8 +958,13 @@ int lcfs_node_add_child(struct lcfs_node_s *parent, struct lcfs_node_s *child,
 		else
 			new_capacity = parent->children_capacity * 2;
 
+#ifdef HAVE_REALLOCARRAY
```
cgwalters, line 961: So this is better in that it doesn't cause a safety/security regression on systems that have `reallocarray`, but it is imposing a duplicate code cost on every user.

I'd really prefer we follow what other projects have done and define our own copy of `reallocarray` if it's missing.

cgwalters (changes requested):

libcomposefs/lcfs-writer.c
alexlarsson: Yeah, and that copy should do the multiply overflow check.

alexlarsson:
> I'll try to update the PR.
> 
> By the way, would that mean, that the other 2 calls to `realloc` in composefs are also unsecured and should be replaced by `reallocarray`?
> 
> https://github.com/containers/composefs/blob/9a9cd591fbddd9a18a5c28b2df8b183d1d8b9ebf/libcomposefs/lcfs-writer.c#L1392-L1393

With the new limit of 16bit n_xattrs, this should be fine.

> https://github.com/containers/composefs/blob/9a9cd591fbddd9a18a5c28b2df8b183d1d8b9ebf/tools/mkcomposefs.c#L750

This already checks for overflow of capacity (that is the `buf->capacity < min_capacity` check above it.

alexlarsson:
Pushed an updated version

alexlarsson:
@ffontaine Can you verify that this works for you?

libcomposefs/lcfs-utils.h
```diff
@@ -42,6 +42,20 @@ static inline char *memdup(const char *s, size_t len)
 	return s2;
 }
 
+static inline bool size_multiply_overflow(size_t size, size_t nmemb) {
+	return nmemb != 0 && size > (SIZE_MAX / nmemb);
+}
+
+#ifndef HAVE_REALLOCARRAY
+static inline void *reallocarray_my(void *ptr, size_t nmemb, size_t size)
```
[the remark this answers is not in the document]
alexlarsson: Oops, sorry, leftover from testing

cgwalters (approved):

cgwalters:
Hello @ffontaine there's a request to relicense this repository, please comment on https://github.com/containers/composefs/issues/344

PR #234, opened 2023-12-13
Bump version to 1.0.2

alexlarsson:
Changes since 1.0.1:

* Dropped composefs-from-json in tests in favour of using the composefs-info dump format.
* libyajl dependency dropped
* libcomposefs now limits the number of xattrs per file to 64k
* Fixed build against libc without reallocarray
* Performance fixes
* go-md2man is used instead of pandoc for manpages
* Minor fixes to spec file


giuseppe (approved):
LGTM

PR #236, opened 2024-01-15
Allow building without go-md2man

yselkowitz:
The cascade of packages which suddenly do not build on i686 due to the introduction of composefs as a dependency is causing issues.  Instead, allow this to be built without manpages.

This is a follow-up to #229.

/cc @cgwalters

### DIFF
--- a/composefs.spec.in
+++ b/composefs.spec.in
@@ -1,3 +1,7 @@
+%ifarch %{golang_arches}
+%bcond man 1
+%endif
+
 Name:           composefs
 Version:        @VERSION@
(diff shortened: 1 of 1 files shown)


cgwalters:
Needs signoff.

Looks OK to me.

(I still don't think we should carry a copy of the spec file upstream, but we don't need to change it now)

cgwalters (approved):

yselkowitz:
Signed off.  Downstream change: https://src.fedoraproject.org/rpms/composefs/pull-request/3

PR #238, opened 2024-01-23
writer: Handle errors from `read`

cgwalters:
Yet another picture perfect example of a bug that Wouldn't Happen In Rust.

Noticed by an internal static analyzer.


alexlarsson:
lgtm

alexlarsson (approved):

PR #244, opened 2024-01-29
Fix use of new mount API and add a try-verity mount option

alexlarsson:
It turns out that the new append mechanism that was added in linux 6.5 that we used was disabled, and a new mechanism was added in 6.7. This changes the new-mount-api code to use the new approach.

Also, I added a "try verity" option that can be used for example by ostree to support fs-verity in an optional way to protect against accidental modification.  This can be passed in if you know fs-verity was enabled on the files, and will not fail if the kernel doesn't support the overlayfs verity= option. In comparison, for the signed usecase we must fail if verity is not supported.


giuseppe (approved):
LGTM

Issue #252, opened 2024-02-07
Avoid the hard requirement of struct mount_attr/MOUNT_ATTR_IDMAP/struct loop_config/LOOP_CONFIGURE


rborn-tx:
When building the project on a Linux Mint 20.3 (based on Ubuntu Focal) I get the following errors:

```
...
  CC       composefs/libcomposefs/libcomposefs_la-lcfs-writer.lo
  CC       composefs/libcomposefs/libcomposefs_la-lcfs-mount.lo
composefs/libcomposefs/lcfs-mount.c:112:13: warning: ‘struct mount_attr’ declared inside parameter list will not be visible outside of this definition or declaration
  112 |      struct mount_attr *attr, size_t usize)
      |             ^~~~~~~~~~
composefs/libcomposefs/lcfs-mount.c: In function ‘setup_loopback’:
composefs/libcomposefs/lcfs-mount.c:273:9: error: variable ‘loopconfig’ has initializer but incomplete type
  273 |  struct loop_config loopconfig = { 0 };
      |         ^~~~~~~~~~~
composefs/libcomposefs/lcfs-mount.c:273:36: warning: excess elements in struct initializer
  273 |  struct loop_config loopconfig = { 0 };
      |                                    ^
composefs/libcomposefs/lcfs-mount.c:273:36: note: (near initialization for ‘loopconfig’)
composefs/libcomposefs/lcfs-mount.c:273:21: error: storage size of ‘loopconfig’ isn’t known
  273 |  struct loop_config loopconfig = { 0 };
      |                     ^~~~~~~~~~
composefs/libcomposefs/lcfs-mount.c:303:20: error: ‘LOOP_CONFIGURE’ undeclared (first use in this function)
  303 |  if (ioctl(loopfd, LOOP_CONFIGURE, &loopconfig) < 0) {
      |                    ^~~~~~~~~~~~~~
composefs/libcomposefs/lcfs-mount.c:303:20: note: each undeclared identifier is reported only once for each function it appears in
composefs/libcomposefs/lcfs-mount.c:273:21: warning: unused variable ‘loopconfig’ [-Wunused-variable]
  273 |  struct loop_config loopconfig = { 0 };
      |                     ^~~~~~~~~~
composefs/libcomposefs/lcfs-mount.c: In function ‘lcfs_mount_erofs’:
composefs/libcomposefs/lcfs-mount.c:384:10: error: variable ‘attr’ has initializer but incomplete type
  384 |   struct mount_attr attr = {
[...]

PR #253, opened 2024-02-09
Support older linux headers

rborn-tx:
This aims to tackle https://github.com/containers/composefs/issues/252, i.e. allowing the build when macros MOUNT_ATTR_IDMAP and LOOP_CONFIGURE along with related structures are not available in the user API of the Linux kernel.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,6 +29,23 @@ jobs:
         with:
           name: composefs.tar
           path: composefs.tar
+  build-baseline:
+    runs-on: ubuntu-latest
+    name: "Build on Ubuntu Focal"
+    container: ubuntu:focal
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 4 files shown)


cgwalters:
This looks sane, but...do you mind trying out adding a suitably old build environment in the CI configuration too?  Otherwise we're somewhat likely to break this again.  You're effectively aiming to define e.g. "Ubuntu Focal" as a baseline.  Should be pretty trivial to pull that as a container image and build there right?

rborn-tx:
> ...do you mind trying out adding a suitably old build environment in the CI configuration too? Otherwise we're somewhat likely to break this again. You're effectively aiming to define e.g. "Ubuntu Focal" as a baseline. Should be pretty trivial to pull that as a container image and build there right?

Sure! I'll do that. Thanks for the feedback.

cgwalters (approved):
Great, thanks!

PR #254, opened 2024-02-15
man mkcomposefs: detail inlining logic

edbaunton:
Whilst playing with mkcomposefs using small dummy files, it's confusing to find that objects are not created in the digest-store. There is an optimization to inline these files in the metadata (see lcfs-internal.h). Add a clarification to help.


cgwalters (approved):
Thank you!

PR #260, opened 2024-03-19
erofs: Handle error from lcfs_build_node_from_image()

eriksjolund:
Adding checks for errors in  lcfs_build_node_from_image() return value.

(I marked this as a draft because I haven't traversed the different code paths yet)

### DIFF
--- a/libcomposefs/lcfs-writer-erofs.c
+++ b/libcomposefs/lcfs-writer-erofs.c
@@ -1452,8 +1452,10 @@ static const erofs_inode *lcfs_image_get_erofs_inode(struct lcfs_image_data *dat
 {
 	const uint8_t *inode_data = data->erofs_metadata + (nid << EROFS_ISLOTBITS);
 
-	if (inode_data >= data->erofs_metadata_end)
+	if (inode_data >= data->erofs_metadata_end) {
+		errno = EINVAL;
 		return NULL;
+	}
 
 	return (const erofs_inode *)inode_data;
 }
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Looks correct to me.

eriksjolund:
I've now read through the different code paths. Looks correct to me too.

cgwalters (approved):
Looks generally right, however scanning things I see we may be missing some errno preservation in callers.

cgwalters:
Hmm, unit tests are hitting
> `AddressSanitizer:DEADLYSIGNAL`
and hanging, which is not a good sign...but not sure it's related.

alexlarsson:
This looks right to me to. And it seems the second unit test passed. However, I'm unable to open the failing first test logs, it just keeps spinning when I expand the "Unit test" section...

@cgwalters Where did you see the errno preservation issues?

I wish there was some kind of gcc attribute for "sets errno".

alexlarsson:
Ah, eventually it opened the log with a ton of DEADLYSIGNAL. I've never seen that before though, seems like something weird.

alexlarsson:
I did some manual tests with asan, as well as with valgrind, and I got nothing. And anyway, its unlikely that any code in this PR got executed it the tests anyway (as its only changes error codepaths), so I doubt it was related. Lets get this in.

alexlarsson:
I wonder if the signal is just ASAN blocking the kill signal that the test gets from the test being cancelled? (For a new run)

PR #262, opened 2024-03-19
handle NULL from lcfs_node_new() and move errno code into the function

eriksjolund:
The interesting commit edits
tools/mkcomposefs.c

The other commit is just an attempt to reduce the number of source code lines.


eriksjolund:
If you want I can remove the part of the PR that rearranges the code.

cgwalters (approved):

PR #263, opened 2024-03-19
writer: Set errno

eriksjolund:
According to [errno(3)](https://man7.org/linux/man-pages/man3/errno.3.html) a function that succeeds is allowed to change errno.
 What if memdup() succeeds? 

I changed  the code to match how it's written here
https://github.com/containers/composefs/blob/d5ffbeb6844cdbe029f9c10cfa98239e2d6b1570/libcomposefs/lcfs-writer.c#L1472


eriksjolund:
I added another small errno modification and rewrote the commit message.

cgwalters (approved):

cgwalters:
At least systemd has a handy `PRESERVE_ERRNO` macro that I've used in other places too, and I think is a nicer approach.  We could add that here.

PR #265, opened 2024-03-25
mountcomposefs: Handle empty basedir option

eriksjolund:
Fail with an error message if any of the
basedir option paths is an empty string.

### DIFF
--- a/tools/mountcomposefs.c
+++ b/tools/mountcomposefs.c
@@ -216,10 +216,12 @@ int main(int argc, char **argv)
 		if (options.objdirs == NULL)
 			errx(EXIT_FAILURE, "Out of memory\n");
 
-		for (i = 0, str = (char *)opt_basedir;; i++, str = NULL) {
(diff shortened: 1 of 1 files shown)


cgwalters (commented):
Hmmm...I guess this is OK, but couldn't we get the same effect with less code by just checking `if (!*opt_basedir) errx(EXIT_FAILURE, "Invalid empty basedir")` ?

eriksjolund:
An empty string might occur anywhere in the colon-separated list.

cgwalters:
`man strtok_r` here says:

> From the above description, it follows that a sequence of two or more contiguous delimiter bytes in the parsed string is considered to be a single delimiter, and that delimiter bytes at the start or end of the string are ignored.  Put another way: the tokens returned by strtok() are always nonempty strings.  Thus, for example, given the string "aaa;;bbb,", successive calls to strtok() that specify the delimiter string ";," would return the strings "aaa" and "bbb", and then a null pointer.

cgwalters (approved):
Oh sorry I misread things.  Yes, this looks fine.

PR #266, opened 2024-03-25
erofs: Add missing free() and lcfs_node_unref()

eriksjolund:


libcomposefs/lcfs-writer-erofs.c
```diff
@@ -1630,6 +1631,8 @@ static struct lcfs_node_s *lcfs_build_node_from_image(struct lcfs_image_data *da
 	new_ht_entry->nid = nid;
 	new_ht_entry->node = node;
 	if (hash_insert(data->node_hash, new_ht_entry) == NULL) {
+		lcfs_node_unref(node);
```
cgwalters, line 1634: I'm OK with this but I think in general it'd be better to do one of two things:

- Use cleanups, and then set the pointers to NULL afterwards
- Better: Change `hash_insert` to *take ownership* of the values unconditionally, i.e. the values are freed even on failure.  This would look like `hash_insert(data->node_hash, &new_ht_entry)` and `new_ht_entry` would always be set to `NULL`.

cgwalters (approved):

libcomposefs/lcfs-writer-erofs.c
eriksjolund: I went for the first alternative as it's not quite clear to me how to deal with the situation when the entry already exists in the hash.

I now also added an assert() statement for documentation purposes. (I don't know if adding an assert() for documentation purposes is good software practice though)

cgwalters: > I went for the first alternative as it's not quite clear to me how to deal with the situation when the entry already exists in the hash.

A good example is `g_hash_table_insert()` which always takes ownership and frees the input value if it already exists.  It's the same way Rust would handle this - ownership consistently transfers to the callee.

cgwalters (approved):

libcomposefs/lcfs-writer-erofs.c
eriksjolund: Thanks, for the reference. I see what you mean.

PR #267, opened 2024-03-25
mkcomposefs: Add missing free()

eriksjolund:


tools/mkcomposefs.c
```diff
@@ -129,6 +129,7 @@ static char *unescape_string(const char *escaped, size_t escaped_size,
 		if (c == '\\') {
 			if (escaped >= escaped_end) {
 				*err = make_error("No character after escape");
+				free(res);
```
cgwalters, line 132: Let's use `cleanup_free` instead, it's designed for this.  The return path will need to `steal_pointer`.

cgwalters (changes requested):
This is OK but I really prefer using the cleanup attributes as a general rule because it makes error handling *much* more reliable.

eriksjolund:
> This is OK but I really prefer using the cleanup attributes as a general rule because it makes error handling _much_ more reliable.

Ok, I'll give it a try.

cgwalters (approved):
Thanks!  I hope you'd agree this is a less error-prone as code is refactored.

eriksjolund:
Yes, this is better.

PR #269, opened 2024-03-25
Added multithreading support for mkcomposefs

### LINKED ISSUE #249: mkcomposefs: multi-threading support
It would be great, if mkcomposefs would support multi-threading. Processing a lot of files takes a long time, even if my machine has 31 idle CPU threads.

divineaugustine:
This resolves #249 and results in 10x improvement for digest calculation and file copy.

Please refer the below document for details about the issue and the change. I would like to know your thoughts.
[threading.md](https://github.com/containers/composefs/files/14743119/threading.md)

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -606,6 +606,64 @@ static void digest_to_path(const uint8_t *csum, char *buf)
 	buf[j] = '\0';
 }
 
+int lcfs_node_set_from_content(struct lcfs_node_s *node, int dirfd,
+			       const char *fname, int buildflags)
+{
+	if (node == NULL) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	bool compute_digest = (buildflags & LCFS_BUILD_COMPUTE_DIGEST) != 0;
+	bool by_digest = (buildflags & LCFS_BUILD_BY_DIGEST) != 0;
+	bool no_inline = (buildflags & LCFS_BUILD_NO_INLINE) != 0;
+	bool is_zerosized = node->inode.st_size == 0;
+	bool do_digest = !is_zerosized && (compute_digest || by_digest);
+	bool do_inline = !is_zerosized && !no_inline &&
+			 node->inode.st_size <= LCFS_BUILD_INLINE_FILE_SIZE_LIMIT;
+	int r;
+
+	if (do_digest || do_inline) {
+		cleanup_fd int fd = openat(dirfd, fname, O_RDONLY | O_CLOEXEC);
+		if (fd < 0)
+			return -1;
+		if (do_digest) {
+			r = lcfs_node_set_fsverity_from_fd(node, fd);
+			if (r < 0)
+				return -1;
+
+			if (by_digest) {
+				const uint8_t *digest =
+					lcfs_node_get_fsverity_digest(node);
+				char digest_path[LCFS_DIGEST_SIZE * 2 + 2];
+				digest_to_path(digest, digest_path);
+				r = lcfs_node_set_payload(node, digest_path);
+				if (r < 0)
+					return -1;
+
+				/* We just computed digest to get the payoad path */
+				if (!compute_digest)
+					node->digest_set = false;
+			}
+
+			/* In case we re-read below */
+			lseek(fd, 0, SEEK_SET);
+		}
+		if (do_inline) {
+			uint8_t buf[LCFS_BUILD_INLINE_FILE_SIZE_LIMIT];
+
+			r = read_content(fd, node->inode.st_size, buf);
+			if (r < 0)
+				return -1;
+			r = lcfs_node_set_content(node, buf, node->inode.st_size);
+			if (r < 0)
+				return -1;
+		}
+	}
+
+	return 0;
+}
+
 struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
 					     int buildflags)
 {
@@ -643,52 +701,9 @@ struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
 	ret->inode.st_size = sb.st_size;
 
 	if ((sb.st_mode & S_IFMT) == S_IFREG) {
-		bool compute_digest = (buildflags & LCFS_BUILD_COMPUTE_DIGEST) != 0;
-		bool by_digest = (buildflags & LCFS_BUILD_BY_DIGEST) != 0;
-		bool no_inline = (buildflags & LCFS_BUILD_NO_INLINE) != 0;
-		bool is_zerosized = sb.st_size == 0;
-		bool do_digest = !is_zerosized && (compute_digest || by_digest);
-		bool do_inline = !is_zerosized && !no_inline &&
-				 sb.st_size <= LCFS_BUILD_INLINE_FILE_SIZE_LIMIT;
-
-		if (do_digest || do_inline) {
-			cleanup_fd int fd =
-				openat(dirfd, fname, O_RDONLY | O_CLOEXEC);
-			if (fd < 0)
-				return NULL;
-			if (do_digest) {
-				r = lcfs_node_set_fsverity_from_fd(ret, fd);
-				if (r < 0)
-					return NULL;
-
-				if (by_digest) {
-					const uint8_t *digest =
-						lcfs_node_get_fsverity_digest(ret);
-					char digest_path[LCFS_DIGEST_SIZE * 2 + 2];
-					digest_to_path(digest, digest_path);
-					r = lcfs_node_set_payload(ret, digest_path);
-					if (r < 0)
-						return NULL;
-
-					/* We just computed digest to get the payoad path */
-					if (!compute_digest)
-						ret->digest_set = false;
-				}
-
-				/* In case we re-read below */
-				lseek(fd, 0, SEEK_SET);
-			}
-			if (do_inline) {
-				uint8_t buf[LCFS_BUILD_INLINE_FILE_SIZE_LIMIT];
-
-				r = read_content(fd, sb.st_size, buf);
-				if (r < 0)
-					return NULL;
-				r = lcfs_node_set_content(ret, buf, sb.st_size);
-				if (r < 0)
-					return NULL;
-			}
-		}
+		r = lcfs_node_set_from_content(ret, dirfd, fname, buildflags);
+		if (r < 0)
+			return NULL;
 	} else if ((sb.st_mode & S_IFMT) == S_IFLNK) {
 		char target[PATH_MAX + 1];
 
--- a/libcomposefs/lcfs-writer.h
+++ b/libcomposefs/lcfs-writer.h
@@ -154,4 +154,7 @@ LCFS_EXTERN int lcfs_compute_fsverity_from_fd(uint8_t *digest, int fd);
 LCFS_EXTERN int lcfs_compute_fsverity_from_data(uint8_t *digest, uint8_t *data,
 						size_t data_len);
 
+LCFS_EXTERN int lcfs_node_set_from_content(struct lcfs_node_s *node, int dirfd,
+					   const char *fname, int buildflags);
+
 #endif
--- a/tools/mkcomposefs.c
+++ b/tools/mkcomposefs.c
@@ -37,6 +37,8 @@
 #include <sys/ioctl.h>
 #include <linux/fsverity.h>
 #include <linux/fs.h>
+#include <pthread.h>
+#include <sys/sysinfo.h>
 
 static void oom(void)
 {
@@ -65,7 +67,8 @@ static __attribute__((format(printf, 1, 2))) char *make_error(const char *fmt, .
 #define OPT_USER_XATTRS 112
 #define OPT_FROM_FILE 113
 #define OPT_MIN_VERSION 114
-#define OPT_MAX_VERSION 115
+#define OPT_THREADS 115
+#define OPT_MAX_VERSION 116
 
 static size_t split_at(const char **start, size_t *length, char split_char,
 		       bool *partial)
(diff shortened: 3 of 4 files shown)


libcomposefs/lcfs-writer.c
```diff
@@ -471,9 +471,16 @@ struct lcfs_node_s *lcfs_node_new(void)
...
 		return NULL;
 	}
-
+	
+	memset( node, 0, sizeof(struct lcfs_node_s));
+	node->payload = NULL;
+	node->content = NULL;
 	node->ref_count = 1;
 	node->inode.st_nlink = 1;
+	node->delayedDigestCalculation = false;
```
cgwalters, line 480: Naming scheme in this project for member variables is `snake_case`, not `camelCase`.

cgwalters (commented):
Thank you so much for working on this!  This may take some time to review as multithreading and C is tricky.

Can you please run `clang-format`?  Basically let's get past the superficial bits before we can dig into the threading.

cgwalters (dismissed):
Marking requested-changes to start just for formatting/style

libcomposefs/lcfs-writer.c
divineaugustine: I have corrected it here and in rest of my changes.

divineaugustine:
> Can you please run `clang-format`? Basically let's get past the superficial bits before we can dig into the threading.
Thank you for checking. I have run clang-format on my changes via vscode formatting. Is that what you meant? 
FYI: I did rebase and force push to fix DCO error.

divineaugustine:
Sorry for the inconvenience one header file was missed out from formatting. I have corrected that as well. Hopefully this solves all the formatting issues from my changeset.

alexlarsson:
I'll have a look at the code, but can you please rebase this and squash the fixups so the end result is a minimal set of independent changes. Say one for the liibrary changes, and one of the mkcomposefs use of it.

libcomposefs/lcfs-writer.c
```diff
@@ -471,9 +471,16 @@ struct lcfs_node_s *lcfs_node_new(void)
 		errno = ENOMEM;
 		return NULL;
 	}
-
+	
+	memset( node, 0, sizeof(struct lcfs_node_s));
```
alexlarsson: No need to memset, calloc() already does that.

libcomposefs/lcfs-writer.c
```diff
@@ -471,9 +471,16 @@ struct lcfs_node_s *lcfs_node_new(void)
...
+	node->payload = NULL;
+	node->content = NULL;
 	node->ref_count = 1;
 	node->inode.st_nlink = 1;
+	node->delayedDigestCalculation = false;
+	node->delayedInline = false;
+	node->delayedPayload = false;
+	node->delayedResetDigest = false;
```
alexlarsson: Additionally, we generally rely on the zero initializing to do all these. Only non-zero initializers are needed.

libcomposefs/lcfs-writer.c
```diff
@@ -590,7 +635,7 @@ static int read_content(int fd, size_t size, uint8_t *buf)
 	return 0;
 }
 
-static void digest_to_path(const uint8_t *csum, char *buf)
+void lcfs_digest_to_path(const uint8_t *csum, char *buf)
```
alexlarsson, line 639: This is maybe not a great function to make public it is very easy to use to shoot yourself in the foot, as it requires the caller to allocate space and whatnot. I think its better to move to lcfs-utils.h or something so that it can be used in mkcomposefs without needing to be in the public library ABI forever.

libcomposefs/lcfs-writer.h
```diff
@@ -37,6 +37,7 @@ enum {
 	LCFS_BUILD_NO_INLINE = (1 << 4),
 	LCFS_BUILD_USER_XATTRS = (1 << 5), /* Only read user.* xattrs */
 	LCFS_BUILD_BY_DIGEST = (1 << 6), /* Refer to basedir files by fs-verity digest */
+	LCFS_BUILD_DELAYED_DIGEST = (1 << 7), /* Refer to basedir files by fs-verity digest */
```
alexlarsson, line 40: Comment is wrong

libcomposefs/lcfs-writer.c
```diff
@@ -564,7 +571,45 @@ int lcfs_node_set_fsverity_from_fd(struct lcfs_node_s *node, int fd)
...
+{
+	return node->delayedPayload;
+}
+
+bool lcfs_get_delayed_inline(struct lcfs_node_s *node)
+{
+	return node->delayedInline;
+}
+
+bool lcfs_get_reset_digest(struct lcfs_node_s *node)
```
alexlarsson, line 608: Why is this not called get_delayed_reset_digest?

libcomposefs/lcfs-writer.c
```diff
@@ -564,7 +571,45 @@ int lcfs_node_set_fsverity_from_fd(struct lcfs_node_s *node, int fd)
...
+{
+	return node->delayedResetDigest;
+}
+
+int lcfs_read_content(int fd, size_t size, uint8_t *buf)
```
alexlarsson, line 613: This too (like digest_to_path) should be an internal function, not a public library  API that we support forever.

tools/mkcomposefs.c
```diff
@@ -760,46 +762,27 @@ static int mkdir_parents(const char *pathname, int mode)
...
-				continue;
-			return -1;
-		}
+	off_t len, ret;
+	len = stat.st_size;
 
-		if (bytes_read == 0)
+	do 
+	{
+		ret = copy_file_range(sfd, NULL, dfd, NULL, len, 0);
```
alexlarsson, line 776: I like this change, but it should be a separate independent commit with commit messages describing it.

tools/mkcomposefs.c
```diff
@@ -893,6 +876,307 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+	struct lcfs_node_s		*node;
+	struct ParallelCopyInfo	*next;
+	char 					*parentPath;
+};
+
+static struct ParallelCopyInfo* new_parallel_copy_info(void)
+{
+	struct ParallelCopyInfo *ret = 
+		calloc(1, sizeof( struct ParallelCopyInfo));
+	ret->next 		= NULL;
```
alexlarsson: calloc already NULLed these.

tools/mkcomposefs.c
```diff
@@ -893,6 +876,307 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+struct ParallelCopyInfo
+{
+	struct lcfs_node_s		*node;
+	struct ParallelCopyInfo	*next;
+	char 					*parentPath;
+};
+
+static struct ParallelCopyInfo* new_parallel_copy_info(void)
+{
+	struct ParallelCopyInfo *ret =
```
alexlarsson: Need oom() check.

tools/mkcomposefs.c
```diff
@@ -893,6 +876,307 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+	ret->parentPath = NULL;
+
+	return ret;
+}
+
+static void cleanup_copy_node(struct ParallelCopyInfo *node)
+{
+	if(node) 
+	{
+		cleanup_copy_node(node->next);
```
alexlarsson: This is a non-tailcall recursion, that is going to use up a lot of stack.
This should just be a regular loop over the nodes instead.

tools/mkcomposefs.c
```diff
@@ -893,6 +876,307 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+			current = construct_copy_data(lcfs_node_get_child(node, i), 
+											current, path);
+		} 
+	}
+	else if ((lcfs_node_get_mode(node) & S_IFMT) == S_IFREG &&
+		   		lcfs_node_get_content(node) == NULL 		&&
+		   		lcfs_node_get_payload(node) != NULL) 
+	{
+		current->node 		= node;
+		current->parentPath = strdup(path);
```
alexlarsson: check strdup return value (and in other strdup calls)

tools/mkcomposefs.c
```diff
@@ -893,6 +876,307 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+
+		path = tmp_path;
+	}
+
+	if ( lcfs_node_dirp(node)) 
+	{
+		const size_t n_children = lcfs_node_get_n_children(node);
+		for (size_t i = 0; i < n_children; i++)
+		{
+			current = construct_copy_data(lcfs_node_get_child(node, i),
```
alexlarsson: No error checking here.

tools/mkcomposefs.c
```diff
@@ -893,6 +876,307 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+{
+	if(node) 
+	{
+		cleanup_copy_node(node->next);
+		free(node->parentPath);
+		free(node);
+	}	
+}
+
+static struct ParallelCopyInfo* construct_copy_data( struct lcfs_node_s* node,
```
alexlarsson: I don't like how this uses a current argument which is returned changed. This will just cause problems when you want to return an error or something like that. Instead create a state structure for the parallel work and put the current pointer in it, and then pass this state to all functions and modify it, then use the return value for error reporting.

tools/mkcomposefs.c
```diff
@@ -893,6 +876,307 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+							sizeof(pthread_t));
+	for( int i = 0; i< nrThreads; i++) 
+		pthread_create(&threads[i], 
+					NULL, 
+					thread_proc, 
+					&threadInfo);
+
+	for( int i = 0; i< nrThreads; i++) 
+		pthread_join(threads[i], NULL);
+
```
alexlarsson, line 1106: I think this should create n_threads - 1 thread, and then call thread_proc on the main thread as well. Then we can efficiently handle the n_thread == 1 case without creating any threads.

tools/mkcomposefs.c
```diff
@@ -1046,6 +1337,7 @@ int main(int argc, char **argv)
 	long min_version = 0;
 	long max_version = 0;
 	char *end;
+	int threads = 1;
```
alexlarsson, line 1298: I think the default should be n_cpus, nobody is going to use this otherwise.

tools/mkcomposefs.c
```diff
@@ -893,6 +876,307 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
 
+static void log_message_with_time( const char* msg )
+{
+	time_t t 		= time(NULL);
+  	struct tm tm 	= *localtime(&t);
+	fprintf(stdout, "%02d:%02d:%02d\t %s\n", 
+				tm.tm_hour, tm.tm_min, tm.tm_sec, msg);
+}
+
+struct ParallelCopyInfo
```
alexlarsson: Really, i think it makes more sense for this to be an array instead of a linked list. With the change to a state struct to keep the current pointer that should be an easy conversion to make. There shouldn't be a need to allocate a ton of individual ParallelCopyInfo objects.

tools/mkcomposefs.c
```diff
@@ -893,6 +876,307 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+		r = lcfs_read_content(fd, sb.st_size, buf);
+		if (r < 0)
+			return r;
+
+		r = lcfs_node_set_content(ret, buf, sb.st_size);
+		if (r < 0)
+			return r;
+	}
+
+	return 0;
```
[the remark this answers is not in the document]
alexlarsson: In fact, I don't think you even need LCFS_BUILD_DELAYED_DIGEST. Just strip out LCFS_BUILD_BY_DIGEST 
 and LCFS_BUILD_COMPUTE_DIGEST from the call to lcfs_load_node_from_file()  and always pass NO_INLINE.
That will have the same effect as DELAYED_DIGEST, and then you just call set_from_content() with the original build flags.

alexlarsson:
So, lots of comments. Make sure you read them all first, because some will make the other ones perhaps not needed. In particular, the propsed change to lcfs_node_set_from_content() will make most of the library API changes unnecessary.

libcomposefs/lcfs-writer.c
divineaugustine: That was a miss. These changes are not required anymore

libcomposefs/lcfs-writer.c
divineaugustine: With the latest change, this remains as an internal function.

libcomposefs/lcfs-writer.h
divineaugustine: This is not required anymore

libcomposefs/lcfs-writer.c
divineaugustine: This remains as an internal function as per the new changes.

tools/mkcomposefs.c
divineaugustine: Added a separate commit just for this change

libcomposefs/lcfs-writer.c
divineaugustine: This is not required anymore.

libcomposefs/lcfs-writer.c
divineaugustine: Thank you. These are not required anymore.

tools/mkcomposefs.c
divineaugustine: Changed this to an array instead of linked list. This code is not valid anymore.

tools/mkcomposefs.c
divineaugustine: added checks in all the allocations.

tools/mkcomposefs.c
divineaugustine: Corrected

tools/mkcomposefs.c
```diff
@@ -1181,14 +1481,32 @@ int main(int argc, char **argv)
...
 		root = lcfs_build(AT_FDCWD, src_path, buildflags, &failed_path);
+		log_message_with_time( "lcfs_build <" );	
 		if (root == NULL)
 			err(EXIT_FAILURE, "error accessing %s", failed_path);
 
-		if (digest_store_path &&
-		    fill_store(root, src_path, digest_store_path) < 0)
-			err(EXIT_FAILURE, "cannot fill store");
+		fflush(stdout); 	
+		if(threads > 1)
```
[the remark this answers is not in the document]
divineaugustine: Yes i totally agree with you. mkcomposefs should by default be multi-threaded in my opinion. I wanted to keep the existing flow untouched in this PR and get your feedback about threading before making threads default.

tools/mkcomposefs.c
divineaugustine: I liked this proposal. I changed it.

tools/mkcomposefs.c
divineaugustine: Changed it to array

tools/mkcomposefs.c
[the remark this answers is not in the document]
divineaugustine: Thank you for this suggestion. This now makes the library code simpler. I also had the same thought in my mind about duplication but i was not so sure about the flags.

divineaugustine:
> So, lots of comments. Make sure you read them all first, because some will make the other ones perhaps not needed. In particular, the propsed change to lcfs_node_set_from_content() will make most of the library API changes unnecessary.

Thank you for the comments. I have addressed those and also squashed the commits to only 3 separate ones as you suggested. As i don´t know who should resolve the comments, i just added reply to those.

libcomposefs/lcfs-writer.c
```diff
@@ -606,6 +606,62 @@ static void digest_to_path(const uint8_t *csum, char *buf)
 	buf[j] = '\0';
 }
 
+bool lcfs_node_set_from_content(struct lcfs_node_s *node, int dirfd,
+				const char *fname, int buildflags)
```
alexlarsson, line 610: All other functions return int, where 0 == success and < 0 is error. I think we should keep doing that.

tools/mkcomposefs.c
```diff
@@ -760,46 +760,25 @@ static int mkdir_parents(const char *pathname, int mode)
...
-				continue;
-			return -1;
-		}
+	off_t len, ret;
+	len = stat.st_size;
 
-		if (bytes_read == 0)
+	do {
+		ret = copy_file_range(sfd, NULL, dfd, NULL, len, 0);
+		if (ret == -1)
```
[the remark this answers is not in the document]
alexlarsson: I wonder also, goes glibc have a fallback for older kernels for this?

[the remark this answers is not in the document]
alexlarsson: Yeah, seems like it.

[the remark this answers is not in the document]
divineaugustine: EINTR is not mentioned as a possible error for copy_file_range. Should that be handled?

libcomposefs/lcfs-writer.c
divineaugustine: fixed

libcomposefs/lcfs-writer.c
```diff
@@ -606,6 +606,62 @@ static void digest_to_path(const uint8_t *csum, char *buf)
...
+{
+	if (node == NULL)
+		return false;
```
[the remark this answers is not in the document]
divineaugustine: fixed

tools/mkcomposefs.c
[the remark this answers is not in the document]
alexlarsson: I'm not sure if the syscall is guaranteed to never get EINTR (although I doubt it), but the glibc fallback implementation clearly doesn't handle it. So, yes, we need to handle EINTR.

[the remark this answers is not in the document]
divineaugustine: Fixed. Thank you

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+		if (new_children == NULL) {
+			errno = ENOMEM;
+			return false;
+		}
+
+		header->items = new_children;
+	}
+	header->items[header->count].node = node;
+	header->items[header->count].path = strdup(path);
+	if (header->items[header->count].path == NULL) {
```
alexlarsson, line 911: Should set errno

alexlarsson (changes requested):
Various nitpicks, but overall this looks good to me.

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
 }
 
-static int fill_store(struct lcfs_node_s *node, const char *path,
-		      const char *digest_store_path)
+struct work_item {
+	struct lcfs_node_s *node;
+	char *path;
+};
+
+struct work_item_header {
```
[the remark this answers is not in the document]
divineaugustine: renamed to work_collection

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+				   struct lcfs_node_s *node, const char *path)
 {
-	cleanup_free char *tmp_path = NULL;
-	const char *fname;
-	int ret;
+	if (!header)
+		return false;
+
+	if (header->count == header->capacity) {
+		header->capacity = header->count == 0 ? 16 : header->capacity * 2;
```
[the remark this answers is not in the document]
divineaugustine: fixed

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+	char *path;
+};
+
+struct work_item_header {
+	struct work_item *items;
+	int count;
+	int capacity;
+};
+
+static bool add_to_work_collection(struct work_item_header *header,
```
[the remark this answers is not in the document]
divineaugustine: fixed

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+		new_children = reallocarray(header->items, sizeof(*header->items),
+					    header->capacity);
+		if (new_children == NULL) {
+			errno = ENOMEM;
+			return false;
+		}
+
+		header->items = new_children;
+	}
+	header->items[header->count].node = node;
```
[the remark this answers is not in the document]
divineaugustine: added ref and unref

tools/mkcomposefs.c
```diff
@@ -1160,12 +1426,20 @@ int main(int argc, char **argv)
 		if (close_input)
 			fclose(input);
 	} else {
+		// Digest calculation and inline will be done in parallel
+		buildflags &= ~LCFS_BUILD_COMPUTE_DIGEST;
+		buildflags &= ~LCFS_BUILD_BY_DIGEST;
+		buildflags |= LCFS_BUILD_NO_INLINE;
```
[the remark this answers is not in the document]
divineaugustine: fixed

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+	thread_info.data = data;
+	thread_info.proc = proc;
+	thread_info.header = header;
+	thread_info.iterator = &iterator;
+
+	pthread_t *threads = NULL;
+	const int thread_count = requested_threads - 1;
+	if (thread_count >= 1) {
+		threads = calloc(thread_count, sizeof(pthread_t));
+		if (threads == NULL)
```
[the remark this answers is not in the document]
divineaugustine: fixed

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+			return false;
+
+		for (int i = 0; i < thread_count; i++) {
+			pthread_create(&threads[i], NULL, thread_proc, &thread_info);
```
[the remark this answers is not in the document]
divineaugustine: fixed

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+		}
+	}
+
+	// Let this thread also process items instead of waiting for the worker threads
+	thread_proc(&thread_info);
+
+	if (thread_count >= 1) {
+		for (int i = 0; i < thread_count; i++)
+			pthread_join(threads[i], NULL);
```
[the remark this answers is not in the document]
divineaugustine: fixed

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+
+	while (true) {
+		struct work_item *item =
+			get_next_work_item(info->header, info->iterator);
+
+		if (!item)
+			return 0;
+
+		if (!item->node)
+			return 0;
```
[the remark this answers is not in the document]
divineaugustine: yes. it is an error if this happens. so i have added a cancel and return in case of empty node.

tools/mkcomposefs.c
```diff
@@ -872,43 +875,283 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+	// Let this thread also process items instead of waiting for the worker threads
+	thread_proc(&thread_info);
+
+	if (thread_count >= 1) {
+		for (int i = 0; i < thread_count; i++)
+			pthread_join(threads[i], NULL);
+
+		free(threads);
+	}
+	return !iterator.cancel_request;
```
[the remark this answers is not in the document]
divineaugustine: request cancel is changed to also send error code. so it is set as errorno before returning from this function.

tools/mkcomposefs.c
divineaugustine: fixed

divineaugustine:
> Various nitpicks, but overall this looks good to me.

Thank you for reviewing it. I have addressed those and pushed the changes.

tools/mkcomposefs.c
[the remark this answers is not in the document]
alexlarsson: These could do:
	collection->items[collection->count].node = lcfs_node_ref(node);
Instead of ref:ing on a separate line

tools/mkcomposefs.c
```diff
@@ -893,43 +877,318 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+	iterator.errorcode = 0;
+	iterator.cancel_request = false;
+
+	struct thread_data thread_info;
+	thread_info.data = data;
+	thread_info.proc = proc;
+	thread_info.collection = collection;
+	thread_info.iterator = &iterator;
+
+	pthread_t *threads = NULL;
```
alexlarsson, line 1109: I would use:
cleanup_free phread_t *threads = NULL;

then you can drop all calls to free(threads)

tools/mkcomposefs.c
```diff
@@ -893,43 +877,318 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+	iterator->cancel_request = true;
+	iterator->errorcode = errorcode;
+	pthread_mutex_unlock(&(iterator->mutex_node_iterator));
+}
+
+typedef int (*THREAD_PROCESS_PROC)(struct work_item *, void *);
+
+static int process_copy(struct work_item *item, void *digest_store_path)
+{
+	if (copy_file_with_dirs_if_needed(item->path, (const char *)digest_store_path,
```
alexlarsson, line 1043: I don't think you need the if here. You can just: "return copy_file_with_dirs_if_needed(...);"

alexlarsson:
Some more minor feedback, but I would like @cgwalters or @giuseppe to also review this.

alexlarsson:
Also, if you're able to talk about it, I'd love to hear what you're using composefs for. Its good to know what you're users are doing.

tools/mkcomposefs.c
divineaugustine: that was a miss. fixed

tools/mkcomposefs.c
```diff
@@ -893,43 +877,318 @@ static ssize_t write_cb(void *_file, void *buf, size_t count)
...
+		iterator->current_item++;
+	}
+	pthread_mutex_unlock(&(iterator->mutex_node_iterator));
+	return cancel ? NULL : ret;
+}
+
+static void request_cancel(struct work_item_iterator *iterator, int errorcode)
+{
+	pthread_mutex_lock(&(iterator->mutex_node_iterator));
+	iterator->cancel_request = true;
```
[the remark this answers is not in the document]
divineaugustine: fixed

tools/mkcomposefs.c
divineaugustine: fixed

tools/mkcomposefs.c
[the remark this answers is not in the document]
divineaugustine: fixed

divineaugustine:
> Also, if you're able to talk about it, I'd love to hear what you're using composefs for. Its good to know what your users are doing.

 Can you give some details @https://github.com/r0l1 ? He is the right person to comment on this.

tools/mkcomposefs.c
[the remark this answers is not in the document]
giuseppe: I think we still need a fallback to the older code, since `copy_file_range` doesn't seem to work across different file systems:

```
$ ./tools/mkcomposefs --digest-store=/tmp/store  tests /dev/null
mkcomposefs: cannot fill store: Invalid cross-device link
```

giuseppe (commented):
left a comment, could you please sign the commits with your real name?

tools/mkcomposefs.c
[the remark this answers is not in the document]
divineaugustine: Do you suggest checking for error code ```EXDEV``` and fallback to previous implementation of read and write or totally replacing copy_file_range with read/write? copy_file_range is already a fallback for soft link attempt. Overall performance gain in my case was ~1% with this api change(i didn't run a lot of bench marks just for this).

divineaugustine:
> left a comment, could you please sign the commits with your real name?

My real name is Divin and i have signed off the commits with it. Did you mean including initials Divin OA or even initials expanded Divin Ookken Athappan?

giuseppe:
> My real name is Divin and i have signed off the commits with it. Did you mean including initials Divin OA or even initials expanded Divin Ookken Athappan?

yes please use the expanded version

tools/mkcomposefs.c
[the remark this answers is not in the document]
giuseppe: > copy_file_range is already a fallback for soft link attempt.

that still fails with different file systems; and I think it is good to keep the new code as it is faster.

I don't think we can add the fallback for any error as the fds could be already used, but `EXDEV` looks safe.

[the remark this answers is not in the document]
cgwalters: It's worth using existing codebases as a reference for which errno values to use as fallback:

- Rust: https://github.com/rust-lang/rust/blob/0e5f5207881066973486e6a480fa46cfa22947e9/library/std/src/sys/pal/unix/kernel_copy.rs#L622
- libglnx: https://gitlab.gnome.org/GNOME/libglnx/-/blob/202b294e6079e23242e65e0426f8639841d1210b/glnx-fdio.c#L846 (which was long ago forked from systemd):
- systemd: https://github.com/systemd/systemd/blob/e71b40fd0026c0884ca26eb4f0a9fbe4d9285cfa/src/shared/copy.c#L338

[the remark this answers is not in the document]
divineaugustine: Thank you for the links. Reading through https://lwn.net/Articles/846403/ i think it is safer to rollback `copy_file_range` to older `read/write` or perhaps address this as a separate PR as this does not have a direct dependency to the original multi-threading change. What do you say? I am very new to linux and i was looking for better apis to copy a file without user mode buffer.

cgwalters:
Yes, separate PRs for things like this that are logically separate please.

> Thank you for the links. Reading through 
> https://lwn.net/Articles/846403/ i think it is safer to rollback 
> `copy_file_range` to older `read/write` or perhaps address this as a 
> separate PR as this does not have a direct dependency to the original 
> multi-threading change. What do you say? I am very new to linux and i 
> was looking for better apis to copy a file without user mode buffer.

divineaugustine:
> Yes, separate PRs for things like this that are logically separate please.

I will create a separate PR for it. Should i also squash up the current commits in this PR to just 2 ?

cgwalters:
Yes, I think one or two commits here is best.

cgwalters:
I'm assuming your goal is faster `mkcomposefs` overall (on a server side build system, right?).  Using multiple threads is one way to do that.  However, another way is caching - ostree does this thing where the build system can retain a cache of (device, inode number) :arrow_right:  checksum; in theory we could offer the same.

The downside is of course that if something happens to mutate a file in the cache, you get incorrect checksums.  But this is pretty easily mitigated, and actually a great thing with composefs is such a failure should be detected at runtime.

(These are complementary approaches to be clear; threading also makes sense)

divineaugustine:
> Yes, I think one or two commits here is best.



> Yes, I think one or two commits here is best.

Squashed the commits to just 2 (one for library and one for mkcomposefs) and siged it off with my full name. Added an issue #274  for `copy_file_range`

alexlarsson:
@r0l1 That sounds cool! Nice that it works for you.

divineaugustine:
> I'm assuming your goal is faster `mkcomposefs` overall (on a server side build system, right?). Using multiple threads is one way to do that. However, another way is caching - ostree does this thing where the build system can retain a cache of (device, inode number) ➡️ checksum; in theory we could offer the same.
> 
> The downside is of course that if something happens to mutate a file in the cache, you get incorrect checksums. But this is pretty easily mitigated, and actually a great thing with composefs is such a failure should be detected at runtime.
> 
> (These are complementary approaches to be clear; threading also makes sense)

Thank you for these alternative-approaches. My objective was to make mkcomposefs run fast in any machine as part of our build process. It was simple and straight forward for me to add threads in mkcomposefs which does not require any facilitation from the end user. I am new to linux tooling and I would require some consultations here to fully understand these alternative approaches.

alexlarsson:
> I'm assuming your goal is faster `mkcomposefs` overall (on a server side build system, right?). Using multiple threads is one way to do that. However, another way is caching - ostree does this thing where the build system can retain a cache of (device, inode number) ➡️ checksum; in theory we could offer the same.

This only works though if you image compose produces hardlinked copies of files you previously committed (otherwise the device/inode of the new files will not match something old in the cache). This happens regularly in ostree because of how it is set up, but that isn't necessarily always true, and it causes some level of extra pain to enforce when you build using ostree. (With things like rofiles-fuse, etc.)

Not saying it is bad, but its not always applicable.

But anyway, if you *do* use a setup like that where you end up with hardlinked files, then the best way to store and quickly access the cached digest is probably to just enable fs-verity on these files. In other words, we should probably have lcfs_node_set_from_content() start by asking the kernel for the fs-verity digest in case it is set, rather than computing it ourselves.

cgwalters:
> In other words, we should probably have lcfs_node_set_from_content() start by asking the kernel for the fs-verity digest in case it is set, rather than computing it ourselves.

Yes agreed, this is an obvious step.

alexlarsson:
I think this looks good enough now. I looked through it again and did some local tests, seems to work well, and the public library APIs are sane.

alexlarsson (approved):

Issue #272, opened 2024-04-08
composefs fails to build with musl


fboudra:
composefs fails to build when using latest musl:

```
tools/mountcomposefs.c warnings and 2: errors generated.
43:20: error: call to undeclared function 'basename'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   43 |         const char *bin = basename(argv0);
      |                           ^
tools/mountcomposefs.c:43:14: error: incompatible integer to pointer conversion initializing 'const char *' with an expression of type 'int' [-Wint-conversion]
   43 |         const char *bin = basename(argv0);
      | 
```

Details on https://errors.yoctoproject.org/Errors/Details/761688/

For reference: https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7

PR #273, opened 2024-04-09
musl: basename: use portable implementation for basename API

fboudra:
musl has removed the non-prototype declaration of basename from string.h which now results in build errors with newer clang compilers.

Implement GNU basename behavior using strchr which is portable across libcs.

Fixes:
| ../../git/tools/mountcomposefs.c:43:20:
| error: call to undeclared function 'basename'; ISO C99 and later do not | support implicit function declarations [-Wimplicit-function-declaration]
|    43 |         const char *bin = basename(argv0);
|       |                           ^
| ../../git/tools/mountcomposefs.c:43:14:
| error: incompatible integer to pointer conversion initializing 'const char *'
| with an expression of type 'int' [-Wint-conversion]
|    43 |         const char *bin = basename(argv0);
|       |                     ^     ~~~~~~~~~~~~~~~

For reference:
https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7

Closes: https://github.com/containers/composefs/issues/272

### DIFF
--- a/libcomposefs/lcfs-utils.h
+++ b/libcomposefs/lcfs-utils.h
@@ -161,4 +161,10 @@ static inline void *steal_pointer(void *pp)
 /* type safety */
 #define steal_pointer(pp) (0 ? (*(pp)) : (steal_pointer)(pp))
 
+static inline const char *gnu_basename(const char *filename)
+{
+	const char *p = strrchr(filename, '/');
+	return p ? p + 1 : filename;
+}
+
 #endif
--- a/tools/mkcomposefs.c
+++ b/tools/mkcomposefs.c
@@ -945,7 +945,7 @@ static void digest_to_string(const uint8_t *csum, char *buf)
 
 static void usage(const char *argv0)
 {
-	const char *bin = basename(argv0);
+	const char *bin = gnu_basename(argv0);
 	fprintf(stderr,
 		"Usage: %s [OPTIONS] SOURCE IMAGE\n"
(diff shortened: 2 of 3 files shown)


quaresmajose:
maybe just the `#include <libgen.h>` can do the job.

it includes the posix implementation of the `basename`
https://man7.org/linux/man-pages/man0/libgen.h.0p.html

fboudra:
@quaresmajose I tried it initially. It looks like we need to use the gnu implementation in this case, not the posix one (including libgen.h won't be enough). Hence, I ended to propose the approach used in kmod. Pending on the authors to review and see if it's acceptable to them.

quaresmajose:
Ok, understand.
My suggestion was mainly because I saw several patches for this issue and many uses the libgen.h.
Thanks for the clarification.

alexlarsson:
Looks fine to me, I'm not overly excited about carrying re-implementations like this, but its trivial enough to not be worth caring about.

alexlarsson:
Ah, seems to be some formating issues, can you run "make clang-format" and repush?

fboudra:
@alexlarsson thanks. formating issue fixed and repushed.

alexlarsson (approved):

cgwalters:
Hello @fboudra there's a request to relicense this repository, please comment on https://github.com/containers/composefs/issues/344

PR #277, opened 2024-04-18
mkcomposefs: Drop newline from error string

alexlarsson:
This will be added when printing the error anyway.


cgwalters:
Missing DCO

giuseppe (approved):

PR #279, opened 2024-05-12
lcfs-fsverity: Add _GNU_SOURCE feature test define

jluebbe:
This file uses htole64(), so it should define _GNU_SOURCE.


alexlarsson:
Its weird that the clang formatter complained about the formatting in a non-changed file. Did clang change?

cgwalters:
> Did clang change?

Yes.  xref https://github.com/ostreedev/ostree/pull/3236 e.g.

cgwalters (approved):

cgwalters:
Will fix up the clang-formatter in a followup

PR #282, opened 2024-05-24
Add a composefs Rust crate

cgwalters:
Right now this includes just a parser/generator for dump files, which is helpful for e.g. a flow like taking a tarball, parsing it in Rust, and then serializing the entry to a composefs dumpfile line.

I'd like to add more opinionated and advanced functionality related to composefs implemented in this crate - *especially* integration with containers/OCI.  My thought is that in the short term we can publish this crate without having the core
C code depend on it.

I plan to use this crate in bootc.

However, more medium term, it may make sense to offer a CLI tool from this crate, in which case there will be some tension with the CLI tools that happen to be implemented in C, as commonly people would

### DIFF
--- a/rust/src/dumpfile.rs
+++ b/rust/src/dumpfile.rs
@@ -1,0 +1,446 @@
+//! # Parsing and generating composefs dump file entry
+//!
+//! The composefs project defines a "dump file" which is a textual
+//! serializion of the metadata file.  This module supports parsing
+//! and generating dump file entries.
+use std::borrow::Cow;
+use std::ffi::OsStr;
+use std::ffi::OsString;
+use std::fmt::Display;
+use std::fmt::Write as WriteFmt;
+use std::io::Write;
+use std::os::unix::ffi::{OsStrExt, OsStringExt};
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+
+use anyhow::Context;
+use anyhow::{anyhow, Result};
+
+#[derive(Debug, PartialEq, Eq)]
+/// An extended attribute entry
+pub struct Xattr<'k> {
+    /// key
+    pub key: Cow<'k, OsStr>,
+    /// value
+    pub value: Cow<'k, [u8]>,
+}
+/// A full set of extended attributes
+pub type Xattrs<'k> = Vec<Xattr<'k>>;
+
+/// Modification time
+#[derive(Debug, PartialEq, Eq)]
+pub struct Mtime {
+    /// Seconds
+    pub sec: u64,
+    /// Nanoseconds
+    pub nsec: u64,
+}
+
+/// A composefs dumpfile entry
+#[derive(Debug, PartialEq, Eq)]
+pub struct Entry<'p> {
+    /// The filename
+    pub path: Cow<'p, Path>,
+    /// uid
+    pub uid: u32,
+    /// gid
+    pub gid: u32,
+    /// mode (includes file type)
+    pub mode: u32,
+    /// Modification time
+    pub mtime: Mtime,
+    /// The specific file/directory data
+    pub item: Item<'p>,
+    /// Extended attributes
+    pub xattrs: Xattrs<'p>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+/// A serializable composefs entry.
+///
+/// The `Display` implementation for this type is defined to serialize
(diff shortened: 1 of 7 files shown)


rust/src/dumpfile.rs
```diff
@@ -0,0 +1,391 @@
...
+            gid,
+            mode,
+            mtime,
+            item,
+            xattrs,
+        })
+    }
+
+    /// Remove internal entries
+    /// FIXME: This is arguably a composefs-info dump bug?
```
cgwalters, line 276: cc @alexlarsson right?  I think these xattrs are internal implementation details and should not leak back into `composefs-info dump` right?

rust/src/mkcomposefs.rs
```diff
@@ -0,0 +1,114 @@
...
+            anyhow::bail!("mkcomposefs failed: {}: {stderr}", r.status)
+        }
+        // SAFETY: We shouldn't fail to join the thread
+        writer.join().unwrap()?;
+        anyhow::Ok(())
+    })
+}
+
+#[test]
+fn test_mkcomposefs() -> Result<()> {
```
cgwalters, line 79: I think this test is pretty neat: we're verifying we can round-trip everything through the Rust parser and serialization of the dump format, through `mkcomposefs` into an EROFS, and back out via `composefs-info dump`.

rust/src/dumpfile.rs
```diff
@@ -0,0 +1,391 @@
...
+        if c != '\\' {
+            write!(r, "{c}").unwrap();
+            continue;
+        }
+        let c = it.next().ok_or_else(|| anyhow!("Unterminated escape"))?;
+        let c = match c {
+            '\\' => b'\\',
+            'n' => b'\n',
+            'r' => b'\r',
+            't' => b'\t',
```
alexlarsson, line 126: We also support `\=` to escape =

alexlarsson: Oh, or at least the docs say, but we don't seem too actually do it.

rust/src/dumpfile.rs
```diff
@@ -0,0 +1,391 @@
...
+                let fsverity_digest = fsverity_digest.as_deref().unwrap_or("-");
+                write!(f, "{fsverity_digest}")?;
+            }
+            _ => {
+                write!(f, "- -")?;
+            }
+        }
+        for xattr in self.xattrs.iter() {
+            f.write_char(' ')?;
+            escape(f, xattr.key.as_bytes())?;
```
alexlarsson, line 367: This should also escape `=` in the key, so the split is guaranteed to work.

rust/src/dumpfile.rs
```diff
@@ -0,0 +1,391 @@
...
+            self.item.size(),
+            self.mode,
+            self.item.nlink(),
+            self.uid,
+            self.gid,
+            self.item.rdev(),
+            self.mtime,
+        )?;
+        if let Some(payload) = self.item.payload() {
+            write!(f, "{payload:?} ")?;
```
alexlarsson, line 342: Don't you need to escape here? Also, you need a custom escape handler for if payload is exactly "-" (to avoid it being parsed as None.

rust/src/dumpfile.rs
```diff
@@ -0,0 +1,391 @@
...
+            write!(f, "- ")?;
+        }
+        match &self.item {
+            Item::Regular {
+                fsverity_digest,
+                inline_content,
+                ..
+            } => {
+                if let Some(content) = inline_content {
+                    escape(f, content)?;
```
alexlarsson, line 353: Same here for "-"

rust/src/dumpfile.rs
alexlarsson: Yes, and we should fix this.

rust/src/dumpfile.rs
alexlarsson: We should probably fix the docs here.

rust/src/dumpfile.rs
cgwalters: Good catch (here and elsewhere!) - fixed.

rust/src/dumpfile.rs
cgwalters: Done in https://github.com/containers/composefs/pull/287

alexlarsson (approved):

PR #285, opened 2024-05-24
lib: two new fsverity APIs

cgwalters:
Add API+CLI to get fsverity digests efficiently

- Add an API that first tries to ask the kernel for the fsverity
  digest, and returns it if present, but otherwise does a fallback
  to in-memory computation.  The primary use case here
  is systems which want to canonically index files by their fsverity
  digest (as opposed to e.g. sha256).  If the filesystem in the
  future starts supporting fsverity (as e.g. XFS is slated to do)
  then this allows an efficient in-place upgrade.
- Expose the measurement in the CLI.  `fsverity-utils` isn't
  installed everywhere, and also today composefs makes hard decisions
  (like using sha256) that that project doesn't do.

---

lib: Add a thin public API wrapper for `FS_IOC_ENABLE_VERITY`

The main thing is this helps ensure that other external
software using the library uses the same fsverity parameters.
There's also the aspect that using ioctl() from some non-C
languages is tricky.

---

### DIFF
--- a/libcomposefs/lcfs-internal.h
+++ b/libcomposefs/lcfs-internal.h
@@ -42,6 +42,9 @@ typedef int errint_t;
  */
 #define LCFS_BUILD_INLINE_FILE_SIZE_LIMIT 64
 
+/* What may be returned by the kernel for digests */
+#define MAX_DIGEST_SIZE 64
+
 #define OVERLAY_XATTR_USER_PREFIX "user."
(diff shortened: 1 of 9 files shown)


alexlarsson (approved):

alexlarsson:
Do we possibly want to be able to do the measurement computation in parallel for multiple files?

cgwalters:
I extended this with another small patch too to add a wrapper for the enable ioctl.

Issue #125, opened 2023-05-20
Add shared library/tool for managing backing store files


cgwalters:
I've been thinking more about [the ostree/composefs integration](https://github.com/ostreedev/ostree/pull/2640) and longer term, I think composefs should have its own opinionated management tooling for backing store files and checkouts.

Basically we move the "GC problem" from higher level tools into a shared composefs layer - and that will greatly "thin out" what ostree needs to do, and the same for container/storage type things.  And more generally, it would help drive *unifying* these two things which I think we do want long term.  Related to this, a mounted composefs shouldn't have backing store files deleted underneath it.

Maybe we could get away with having this just be a directory, e.g. `/composefs` (like `/ostree/repo`) or perhaps `/usr/.composefs`.  Call this a `$composefsdir`.

Vaguely thinking perhaps we could have then `$composefsdir/roots.d` with namespaced subdirectories, like `$composefsdir/roots.d/ostree` and `$composefsdir/roots.d/containers`.  Finally there'd be `$composefsdir/files` which would hold the regular files.

Then we'd have a CLI tool like `/usr/libexec/composefsctl gc` that would iterate over all composefs filesystems and GC any unreferenced regular files.  In order to ensure GC doesn't race with addition we'd also need "API" operations like `/usr/libexec/composefsctl add container/foo.composefs` that did locking.  And a corresponding `composefsctl delete`.

alexlarsson:
I've been discussing the backing-files GC issue with @giuseppe quite a bit in the context of containers/storage. And the best approach I've come up with is this:

Suppose you have `/composefs/` like above, you would in it have a layout something like:
```
├── files
│   ├── 00
│   │   └── 1234.file
│   ├── aa
│   │   └── 5678.file
│   └── bc
│       └── abcd.file
└── images
    ├── foo
    │   ├── image.cfs
    │   ├── 00
    │   │   └── 1234.file
    │   └── aa
    │       └── 5678.file
    └── bar
        ├── image.cfs
        ├── 00
        │   └── 1234.file
        └── bc
            └── abcd.file
```

So, a shared backing file dir with *all* files from all images, and then each image has a  directory with only the files for that image. However, the backing files would be hardlinked. Then, to remove an image you delete the image file and the directory, and then you make a pass over the shared dir and delete any files with n_link==1.

[...]

PR #286, opened 2024-05-26
rust: Add a composefs-oci crate

cgwalters:
The high level goal of this crate is to be an opinionated
generic storage layer using composefs, with direct support
for OCI.  Note not just OCI *containers* but also including
OCI artifacts too.

This crate could to be the successor to
the "storage core" of both ostree and containers/storage.

---

This is an initial sketch!  It'd be good to sync on the design/goals; it started to "feel" right but obviously there's a whole lot going on.  In particular, when I was going through the current containers/storage composefs code...there's a few things here where I am not sure it has the right architecture.

- I think relying on zstd:chunked and especially the part about indexing objects by their sha256 is wrong.  We should index by the fsverity digest - especially in a time when people want to store potentially multi-GB files in container images, computing a checksum in two different ways is quite suboptimal.  Note that using a new C API exposed by this project, if the filesystem doesn't physically support fsverity, we just compute it in userspace.
- The repo also tracks whether fsverity was supported at init time and can be configured to hard require it.  By tracking enablement we can also support a "migrate repo to enabling fsverity" for e.g. XFS when it gains that support.
- Related to the above on the bootc side we definitely want file sharing between images; this will lead to a need for GC handling (xref https://github.com/containers/composefs/issues/125 )
- There's a logistical issue in Go vs Rust going on here...we can bridge it of course, but:
- Managing file paths in userspace makes me nervous; there's lots of subtleties.  This design sketch hard requires and uses openat2 for unpacking, ensuring the kernel handles things like canonicalization of `//` etc. and symlink traversal.

What I want to do next here is teach `mkcomposefs` to optionally honor a `user.cfs.meta` xattr.  This way we can unpack a tarball to disk, but instead of e.g. *physically* setting sensitive file metadata like owner uid, gid, suid bits, and security-related xattrs, we can do something much like what ostree's `bare-user` mode does and store them in an xattr.  We can also skip making specials like device nodes and FIFOs "physically" and just make a zero-sized regular file with the xattr.

### DIFF
--- a/rust/composefs-oci/src/lib.rs
+++ b/rust/composefs-oci/src/lib.rs
@@ -1,0 +1,90 @@
+use std::ffi::OsString;
+
+use anyhow::{Context, Result};
+use camino::Utf8PathBuf;
+use cap_std_ext::dirext::CapStdExtDirExt;
+use clap::Parser;
+use ocidir::cap_std::{
+    self,
+    fs::{Dir, DirBuilder},
+};
+use pull::cli_pull;
+
+mod fileutils;
+pub mod pull;
+pub mod repo;
+
+/// Options for specifying the repository
+#[derive(Debug, Parser)]
+pub(crate) struct RepoOpts {
+    /// Path to the repository
+    #[clap(long, value_parser)]
+    repo: Utf8PathBuf,
+}
+
+impl RepoOpts {
+    pub(crate) fn open(&self) -> Result<crate::repo::Repo> {
+        let repo = self.repo.as_path();
+        let d = Dir::open_ambient_dir(repo, cap_std::ambient_authority())
+            .with_context(|| format!("Opening {repo}"))?;
+        crate::repo::Repo::open(d)
+    }
+}
+
+/// Options for importing container.
+#[derive(Debug, Parser)]
+pub(crate) struct PullOpts {
+    #[clap(flatten)]
+    repo_opts: RepoOpts,
+
+    /// Image reference
+    image: String,
+}
+
+/// Options for creating a repo
+#[derive(Debug, Parser)]
+pub(crate) struct CreateOpts {
+    #[clap(flatten)]
+    repo_opts: RepoOpts,
+
+    /// Require fsverity
+    #[clap(long)]
+    require_verity: bool,
+}
+
+/// Toplevel options
+#[derive(Debug, Parser)]
+#[clap(name = "composefs")]
+#[clap(rename_all = "kebab-case")]
+#[allow(clippy::large_enum_variant)]
+pub(crate) enum Opt {
+    /// Pull an image
+    Create(CreateOpts),
+    /// Pull an image
+    Pull(PullOpts),
+}
+
+/// Parse the provided arguments and execute.
+/// Calls [`clap::Error::exit`] on failure, printing the error message and aborting the program.
+pub async fn run_from_iter<I>(args: I) -> Result<()>
+where
+    I: IntoIterator,
+    I::Item: Into<OsString> + Clone,
+{
+    run_from_opt(Opt::parse_from(args)).await
+}
+
+async fn run_from_opt(opt: Opt) -> Result<()> {
+    match opt {
+        Opt::Create(opts) => {
+            let repopath = opts.repo_opts.repo.as_path();
+            std::fs::create_dir_all(repopath)
+                .with_context(|| format!("Creating target dir: {repopath}"))?;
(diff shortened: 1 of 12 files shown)


rust/composefs-oci/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    #[clap(flatten)]
+    repo_opts: RepoOpts,
+
+    /// Image reference
+    image: String,
+}
+
+/// Toplevel options for extended ostree functionality.
+#[derive(Debug, Parser)]
+#[clap(name = "ostree-ext")]
```
alexlarsson, line 32: ostree-ext?

rust/composefs-oci/src/repo.rs
```diff
@@ -0,0 +1,439 @@
...
+}
+
+impl Repo {
+    pub fn init(fd: Dir, require_verity: bool) -> Result<Self> {
+        let supports_verity = test_fsverity_in(&fd)?;
+        if require_verity && !supports_verity {
+            anyhow::bail!("Requested fsverity, but target does not support it");
+        }
+        let meta = RepoMetadata {
+            version: String::from("0.5"),
```
alexlarsson, line 174: You said 0.1 above.

rust/composefs-oci/src/repo.rs
```diff
@@ -0,0 +1,439 @@
...
+        // so in the future it'd be nice to do this is a helper thread.  However
+        // doing so would significantly complicate the flow.
+        if self.has_verity() {
+            fileutils::reopen_tmpfile_ro(&mut tmpfile).context("Reopening tmpfile")?;
+            composefs::fsverity::fsverity_enable(tmpfile.as_file().as_fd())
+                .context("Failed to enable fsverity")?;
+        };
+        let mut digest = Digest::new();
+        composefs::fsverity::fsverity_digest_from_fd(tmpfile.as_file().as_fd(), &mut digest)
+            .context("Computing fsverity digest")?;
```
alexlarsson, line 336: In the case where the filesystem doesn't support fs-verity, this will re-read the file to compute the fs-verity. We already have fs-verity digest support in libcomposefs that supports streaming, so we could have computed this digest while writing to the tmpfile above.

I guess that would be unnecessary in the case where the target fs supports fs-verity, as then we compute it twice, but we can avoid that if self.has_verity().

alexlarsson:
> What I want to do next here is teach mkcomposefs to optionally honor a user.cfs.meta xattr. This way we can unpack a tarball to disk, but instead of e.g. physically setting sensitive file metadata like owner uid, gid, suid bits, and security-related xattrs, we can do something much like what ostree's bare-user mode does and store them in an xattr. We can also skip making specials like device nodes and FIFOs "physically" and just make a zero-sized regular file with the xattr.

I'm not sure I understand why you want this? During the import, can't you just compute the dump-format file line by line in combination with the object files and pass that to mkcomposefs? Why do you need the file metadata "on-disk"?

rust/composefs-oci/src/repo.rs
cgwalters: It's a good point.  I think though a better optimization will be to move the fsverity computation (whether userspace or kernel) into a worker thread pool, so we leave the main thread's job to basically reading from the network and saving data as quickly as possible.  This way we avoid intermixing network and CPU bound work on the same thread.  (OTOH on modern processors with sha256 acceleration maybe it doesn't matter, I haven't measured)

alexlarsson: Honestly, I don't know what the cost is of sha256 these days. Sometimes modern CPUs are fast enough that doing computation on memory is basically free compared to the memory bandwidth cost, but I don't know if that is true with sha256. And in this case its even compared to i/O as well. If the file is large, then we may actually have to read it back from disk (rather than cache).

I did some tests with sha256sum on my (pretty fast) machine on a 1GB file. cat to /dev/null takes ~270 msec uncached, and ~65 msec cached (real time).  Running sha256sum instead takes 460 msec uncached (although this time varies more), and 440 msec cached. 

I guess we can say that sha256 computation is the more costly operation compared to disk i/o. Cached read+sha is 7x time slower than a cached read. And total sha256 time is almost the same for uncached and cached reads, meaning that the disk read times were small (or at least pipelined) compared to the computation time.

Given the above we might want something like a threadpool that does fixed block size sha56 computation across n_cpu threads. Then we could just feed it block by block, getting parallelization even across a single large file.

cgwalters:
For now I think it makes sense actually to keep this repository with just "core" functionality.
I have created https://github.com/cgwalters/composefs-oci as an experimental place to iterate.

Issue #198, opened 2023-09-22
Game out a plan for a 1.1 format


cgwalters:
Let's assume 1.0 is released, and we discover something like a notable performance issue with the 1.0 format.  Or maybe it's actually broken in an important corner case on big-endian (s390x) - something like that.

Say this is important enough to do a 1.1.

I think the way this would need to work is basically we add support for e.g. `--format=1.1` to the CLI/API - and then we generate *both* digests. 

We need to think through and verify a scenario like this would work:

- Build server is updated with composefs 1.1 support
- Client system (e.g. ostree/container tooling/RAUC/whatever) only has 1.0 support
- Client fetches metadata (OCI image, whatever) that has both digests
- Client ignores the 1.1 digest, synthesizes a 1.0 format, and successfully verifies it against the 1.0 digest
- Client later than updates to tooling (podman/ostree/RAUC) that supports 1.1
- Client synthesizes a 1.1 EROFS, and verifies using that digest

Right?

cgwalters:
It may actually be worth adding a stub 1.1 format *now* that has a trivial change as a hidden option just to really test things out.

alexlarsson:
There is already a `uint32_t version` in `struct lcfs_write_options_s` for this particular reason. But you're right, we should plumb it to the CLI and actually test it.

alexlarsson:
If we wanted a stupid optional feature we could have one that skips the 00-ff whiteouts in the image. That means its only going to work well (i.e. the basedir would not be visible) with kernels that have data-only overlayfs layers, but for those it would be more efficient.

allisonkarlitskaya:
Other potential wishlist item for a trivial change to make things more efficient: more aggressive list of xattr prefixes.  We should really have "prefixes" for the complete length of all of the overlayfs xattrs we output.

alexlarsson:
The use of custom prefixes would be nice, but it does bump up the kernel requirements to 6.4.

allisonkarlitskaya:
Having implemented a second erofs writer, this is something like my list of proposed changes for composefs erofs v1.1 file format:

 - no compact inodes
   - store a 1970 superblock mtime; or
   - use the root dir mtime; or
   - use the newest file?
 - set the 32-bit compatible `ino` field equal to the 64-bit nid value
 - regular files and symlinks can only be inlined
   - limit these to 2k (or 1k)
   - we use blocks only for directories
 - very clearly specify inode alignment-adjustment algorithm for ensuring
[...]

PR #288, opened 2024-05-28
Info drop internal xattrs

cgwalters:
@alexlarsson I took a pass at https://github.com/containers/composefs/pull/282#discussion_r1615667462 (this is only 76.3% baked, I got a little lost a few times in the quoting/processing here) but I'm pretty sure now what we've been writing into the EROFS blob is wrong.  If I'm right about that...I think there's two options:

- Try to roll out our [format version 1.1 game plan](https://github.com/containers/composefs/issues/198)
- Just ignore it...how much do we really care about the `user.` namespace given that we can't mount an EROFS unprivileged anyways and that's super unlikely to change in the future?

### DIFF
--- a/libcomposefs/lcfs-internal.h
+++ b/libcomposefs/lcfs-internal.h
@@ -48,25 +48,39 @@ typedef int errint_t;
 #define OVERLAY_XATTR_USER_PREFIX "user."
 #define OVERLAY_XATTR_TRUSTED_PREFIX "trusted."
 #define OVERLAY_XATTR_PARTIAL_PREFIX "overlay."
+// trusted.overlay.
 #define OVERLAY_XATTR_PREFIX                                                   \
 	OVERLAY_XATTR_TRUSTED_PREFIX OVERLAY_XATTR_PARTIAL_PREFIX
+// user.overlay.
 #define OVERLAY_XATTR_USERXATTR_PREFIX                                         \
 	OVERLAY_XATTR_USER_PREFIX OVERLAY_XATTR_PARTIAL_PREFIX
+// trusted.overlay.overlay.
 #define OVERLAY_XATTR_ESCAPE_PREFIX OVERLAY_XATTR_PREFIX "overlay."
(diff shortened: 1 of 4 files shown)


alexlarsson:
I'm not sure we're actually wrong here. But these things are mind-bending, so maybe I'm mistaken.

Here is the situation this is targeting: We have an "image" that contains a traditional whiteout node. We want to create an erofs+overlayfs from this, such that when it is mounted (by root), the resulting file in the mount:
a) Is not swallowed by the cfs overlayfs mount
b) Is usable as a whiteout in a rootful overlayfs where the lowerdir is on the cfs mount
c) Is usable as a whiteout in a userxattr overlayfs where the lowerdir is on the cfs mount

Now, lets focus on c. In the erofs image we create a regular file with the "user.overlay.whiteout" xattr set. Since the composefs overlayfs is rootful, this xattr is passed on as-is and works as a xwhiteout for a second userxattr mount. However, it needs the parent directory to be marked for overlayfs to handle it. So, we set "user.overlay.opaque" to "x" to mark this. This xattr also is not touched by the composefs overlay mount, so it is available as-is for the second level userxattr overlayfs mount.

I think maybe you misunderstood the reason for these files. They are not because we want to mount the erofs image with userxattrs (that will not work for many reasons), but rather that the rootful overlayfs mount of the erofs image should produce a mount where the whiteout file is useful both for regular and userxattr mounts.

alexlarsson:
Question though, should we convert an xwhiteout to a regular whiteout when we dump?

cgwalters:
> but rather that the rootful overlayfs mount of the erofs image should produce a mount where the whiteout file is useful both for regular and userxattr mounts.

Right!  That makes sense.

cgwalters:
The first commit got reimplemented a different way.

I pushed the second commit to https://github.com/containers/composefs/pull/384

This PR is going to need some nontrivial thought at some point, but isn't critical.

PR #295, opened 2024-06-03
rust: Add mandatory `description` field, other misc changes

cgwalters:
I want to publish `composefs-sys` but a `description` field is required, which is annoyingly not found by `cargo publish --dry-run`.

Do a few other tweaks while we're here, including talking about language bindings in `README.md`.


jeckersb (approved):

cgwalters:
Cool, published https://crates.io/crates/composefs-sys but then that let me hit the next problem in https://github.com/containers/composefs/pull/297

PR #296, opened 2024-06-04
test: remove hardcoded /tmp

Kiskae:
Just upstreaming some things I ran into while I was originally packaging for nixpkgs.
Since the builds happen in a minimal sandbox it was running into non-existing directories.

The only thing I'm not sure about is the tests that involve fsverity since usually `/tmp` is a `tmpfs` and might not support it. But in that case it is probably a better idea to specifically set `$TMPDIR` for those tests rather than hardcoding `/var/tmp`

### DIFF
--- a/tests/test-lib.sh
+++ b/tests/test-lib.sh
@@ -24,7 +24,7 @@ assert_file_has_content () {
 }
 
 check_whiteout () {
-    tmpfile=$(mktemp /tmp/lcfs-whiteout.XXXXXX)
+    tmpfile=$(mktemp --tmpdir lcfs-whiteout.XXXXXX)
     rm -f $tmpfile
     if mknod $tmpfile c 0 0 &> /dev/null; then
         echo y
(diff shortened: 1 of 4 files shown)


tests/test-lib.sh
```diff
@@ -56,7 +56,7 @@ check_erofs_fsck () {
 
 check_fsverity () {
     fsverity --version >/dev/null 2>&1 || return 1
-    tmpfile=$(mktemp /var/tmp/lcfs-fsverity.XXXXXX)
+    tmpfile=$(mktemp --tmpdir lcfs-fsverity.XXXXXX)
```
Kiskae, line 59: Specifically this, its probably a better idea to override `$TMPDIR` before `test-lib.sh` is loaded to get fsverify support.

Maybe something like:

```bash
if [[ -z "${COMPOSEFS_FSVERITY_USE_TMPDIR:-}" ]]; then
    TMPDIR="/var/tmp"
fi
```

cgwalters (approved):

PR #298, opened 2024-06-04
rust: Two minor patches

cgwalters:
rust: Clarify that "composefs-core" is really "composefs"

I originally wanted to make a "core" crate as distinguished
from the higher level potential "composefs-oci" crate,
but the crate name is actually "composefs" so let's clarify that.

---

rust: Add README.md files, link to crates.io

They're now published, so link to crates.io.  Add a `README.md`
for each one too so that there's something on the crates page.

---


giuseppe (approved):
LGTM

PR #299, opened 2024-06-05
composefs-info: set locale to C

giuseppe:
composefs-info is using internally isgraph and isprint that are locale dependent.  Force the current locale to be C .


cgwalters (commented):
I'm not opposed to this but IMO the most sustainable fix is to use explicit ascii versions, e.g. we could copy paste https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gstrfuncs.h#L79

Actually, https://github.com/util-linux/util-linux/blob/d7282ef8b3b030bbecee4264ae7674e0bf08d52c/include/cctype.h#L182 looks very copy-pasteable.

(Or of course, consider using Rust for these tools...)

giuseppe:
would you prefer a version with `isgraph_l` and `isprint_l`?

giuseppe:
switched to isgraph_l and isprint_l

cgwalters (approved):

PR #300, opened 2024-06-26
Add meson as an alternative to autotools with corresponding CI

jluebbe:
This adds support for meson as an alternative build system to autotools.

### DIFF
--- a/libcomposefs/meson.build
+++ b/libcomposefs/meson.build
@@ -1,0 +1,41 @@
+source_files = files([
+  'bitrotate.h',
+  'erofs_fs.h',
+  'erofs_fs_wrapper.h',
+  'hash.c',
+  'hash.h',
+  'lcfs-internal.h',
+  'lcfs-erofs.h',
+  'lcfs-erofs-internal.h',
+  'lcfs-fsverity.c',
+  'lcfs-fsverity.h',
+  'lcfs-writer-erofs.c',
+  'lcfs-writer.c',
+  'lcfs-writer.h',
+  'lcfs-utils.h',
+  'lcfs-mount.c',
+  'lcfs-mount.h',
+  'xalloc-oversized.h',
+])
+
+libcomposefs = both_libraries('composefs',
+  source_files,
+  c_args : composefs_hash_cflags + hidden_visibility_cflags,
+  dependencies : libcrypto_dep,
+  version : libversion,
+  soversion : soversion,
+  include_directories : config_inc,
+  install : true,
+)
+
+pkg.generate(libcomposefs,
+  description : 'library for generating and using composefs images',
+)
+
+install_headers([
+    'lcfs-writer.h',
+    'lcfs-erofs.h',
+    'lcfs-mount.h',
+  ],
+  subdir : 'libcomposefs',
+)
--- a/tools/meson.build
+++ b/tools/meson.build
@@ -1,0 +1,37 @@
+libcomposefs_dep = declare_dependency(link_with : libcomposefs, include_directories : config_inc)
+
+thread_dep = dependency('threads')
+
+executable('mkcomposefs',
+    'mkcomposefs.c',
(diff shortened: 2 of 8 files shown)


libcomposefs/meson.build
```diff
@@ -0,0 +1,44 @@
...
+  'lcfs-mount.c',
+  'lcfs-mount.h',
+  'xalloc-oversized.h',
+])
+
+libcomposefs = shared_library('composefs', 
+  source_files,
+  c_args : composefs_hash_cflags + hidden_visibility_cflags,
+  dependencies : libcrypto_dep,
+  version : meson.project_version(), # FIXME: or libversion?
```
cgwalters, line 27: How does this relate to https://github.com/containers/composefs/pull/300/files#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R16 ?

tools/meson.build
```diff
@@ -0,0 +1,37 @@
...
+executable('mount.composefs',
+    'mountcomposefs.c',
+    dependencies : [libcomposefs_dep],
+    install : true,
+    install_dir : get_option('sbindir'),
+)
+
+executable('composefs-info',
+    ['composefs-info.c', '../libcomposefs/hash.c'],
+    c_args : ['-DUSE_OBSTACK=0', '-DTESTING=0', '-DUSE_DIFF_HASH=0'],
```
cgwalters, line 20: We should be able to reference a variable for this to dedup with the top of https://github.com/containers/composefs/pull/300/files#diff-4f12b63e85afe8c8819073d64da906f24d8fe7ad05e0cc1db6fab193413e4410 right?

cgwalters (approved):
Nice! Only superficially skimming this (and without knowing meson really...my path went autotools -> go+rust basically) it looks great! I'd be in favor of killing the autotools build actually.

The thing I'd like to cross-check at least manually (even in CI perhaps) is that the autotools and meson builds result in "identical" outputs...I am not aware of any reason they shouldn't actually be bit-for-bit compatible.

libcomposefs/meson.build
jluebbe: I saw that you have different versions for the project itself and the libcomposefs.so links. The pkgconfig file also includes a version field. Which version should go where? :)

jluebbe: I've changed it to match what autotools produces (project version 1.0.4 in composefs.pc, libversion 1.2.0 for the symlinks).

jluebbe:
I've noticed that autotools builds both a shared and a static library, so I've updated this PR to do that as well.

cgwalters:
Tried this out, it's so much faster than autotools! Nice!

Doing a `diff -uNr` between the two builds, I see:

```
--- /tmp/composefs-autotools/usr/lib64/pkgconfig/composefs.pc   2024-06-28 09:50:22.074577014 -0400
+++ /tmp/composefs-meson/usr/lib64/pkgconfig/composefs.pc       2024-06-28 09:49:34.856786643 -0400
@@ -1,13 +1,10 @@
 prefix=/usr
-exec_prefix=${prefix}
-libdir=/usr/lib64
 includedir=${prefix}/include
+libdir=${prefix}/lib64
 
-Name: Composefs
+Name: composefs
 Description: library for generating and using composefs images
 Version: 1.0.4
-Requires: 
-Requires.private:  libcrypto
+Requires.private: libcrypto
 Libs: -L${libdir} -lcomposefs
-Libs.private: -lcrypto
 Cflags: -I${includedir}
```

Which looks totally fine to me. Digging into the differences via `diffoscope` on the generated binaries and shared libraries, everything I see looks like it's an artifact of either:

- The source paths I used were different, flowing into all the debuginfo
- Some default linker flags are different

But that's all fine.

Thanks so much for doing this! I'm still in favor of dropping autotools; will file a tracker issue where we can debate that.

PR #303, opened 2024-07-15
mount: Consistently use `composefs` as source

cgwalters:
The "source" field for overlayfs isn't strictly meaningful, by default; but it's useful to identify at least the software that created the mount.

Motivated by https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/301 where some QE testing noticed that an updated composefs using the new mount API didn't specify the source at all, so it ends up as `none`, but if we fall back to the legacy one it uses `overlayfs`.

Change both to be `composefs`.


giuseppe (approved):
LGTM

PR #307, opened 2024-07-30
libcomposefs: do not override errno

giuseppe:
do not override the errno reported by the write_cb as we might lose useful errors like ENOSPC.

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -275,9 +275,9 @@ int lcfs_write(struct lcfs_ctx_s *ctx, void *_data, size_t data_len)
 
 	if (ctx->write_cb) {
 		while (data_len > 0) {
+			errno = EIO;
 			ssize_t r = ctx->write_cb(ctx->file, data, data_len);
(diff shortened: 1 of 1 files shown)


libcomposefs/lcfs-writer.c
```diff
@@ -275,9 +275,9 @@ int lcfs_write(struct lcfs_ctx_s *ctx, void *_data, size_t data_len)
 
 	if (ctx->write_cb) {
 		while (data_len > 0) {
+			errno = EIO;
```
cgwalters, line 278: Hmm, I guess we need to do this because we can't rely on callers setting errno...and indeed in our own code which uses `fwrite`, it doesn't?

So...okay, is there *really* no way to get the actual error out of `ferror`?

cgwalters (approved):

libcomposefs/lcfs-writer.c
cgwalters: Oh, I guess `fwrite` is...somewhat expected conventionally to set `errno`, but it's not standardized?

Issue #301, opened 2024-06-28
dropping autotools build


cgwalters:
Since https://github.com/containers/composefs/pull/300 we now have a meson build. I don't know of a strong reason to keep the autotools one around.

This issue tracks the goal of removing it; if you have objections or concerns, please cite them here.

alexlarsson:
The one thing our meson code doesn't do is generate the .spec from spec.in during dist.
It can be done, like so: https://gitlab.com/CentOS/automotive/src/autoinit/-/commit/2d1bbad976e3729faf864c205dd830cd04d7aa68

Otoh, maybe its time to drop the specfile from the tarball instead?

PR #313, opened 2024-08-15
build-sys: Remove autotools

cgwalters:
meson is better, basically.

Closes: https://github.com/containers/composefs/issues/301


alexlarsson:
LGTM.
Can we remove the spec file too, as we're not doing the version replace during dist anyway.

cgwalters:
> Can we remove the spec file too, as we're not doing the version replace during dist anyway.

Hmm...I personally don't like maintaining spec files in upstream (any more than debian/rules or bitbake recipes, etc...). But that said https://packit.dev/ which is used by other container projects seems to prefer this (although I think it can fetch from the relevant distro dist-gits too).

For now I just made a best effort to update it, but I actually tried updating the fedora spec, and I needed to do this FTR:

```
diff --git a/composefs.spec b/composefs.spec
index b638b0a..cda03ca 100644
--- a/composefs.spec
+++ b/composefs.spec
@@ -3,7 +3,7 @@
 %endif
 
 Name:           composefs
-Version:        1.0.4
+Version:        1.0.5
 Release:        2%{?dist}
 Summary:        Tools to handle creating and mounting composefs images
 
@@ -43,21 +43,14 @@ Library files for %{name}.
 
 %prep
 %autosetup -p1
-# for go-md2man patch
-autoreconf -fiv
 
 %build
-%configure \
-           --disable-static \
-%if %{with man}
-           --enable-man \
-%endif
-           --with-fuse
-%make_build
+%meson --default-library=shared -Dfuse=enabled -Dman=enabled 
+%meson_build
 
 %install
-%make_install
-rm -rf %{buildroot}%{_libdir}/libcomposefs.la
+%meson_install
+rm -v $RPM_BUILD_ROOT/%{_libdir}/libcomposefs*.a
 
 %files devel
 %{_includedir}/libcomposefs
```

I also stumbled on the fact that while we tried to do meson in 1.0.5, I built the release with autotools and we didn't dist the meson rules in the tarball...we could fix that of course, but probably better to just pull the trigger on this.

alexlarsson (approved):

PR #315, opened 2024-08-17
lib: Verify mode values

cgwalters:
Today, `lcfs_node_set_mode` happily accepts any `mode_t` aka uint32 value and basically dumps it into the erofs inode. I accidentally stumbled over this when I forgot that a regular file needs `S_IFREG` in the mode bits in the dump file, but at least `fsck.erofs` does notice this.


cgwalters:
This will textually (but not logically) conflict with https://github.com/containers/composefs/pull/314 due to dump file addition; let's merge that first.

alexlarsson (approved):

alexlarsson:
Hmm, this looks good, but seems to break the CI...

cgwalters:
> Hmm, this looks good, but seems to break the CI...

OK yeah, we need to skip checking the mode for hardlinks as it's not meaningful; done.

PR #317, opened 2024-08-19
Two more testing improvements

cgwalters:
tests: Fix extra `-` in should-fail tests

I had an extra `-` at the end here which was being interpreted
as an xattr...It turns out that `mkcomposefs` is liberal here
in what it accepts (string parsing in C stinks) and interprets
a missing `=` to be an xattr with the provided name, and
no value.

In this *specific* case that's an xattr with the name `-`.

The Rust parser is stricter and rejects this (which I think
is a good thing), and this is preparation for doing stronger
verification on the Rust side before we start passing data
into C.

---

rust: Also test we fail to parse the `should-fail-` cases

This way, callers of the Rust API get validation before
we start passing things down to the C code.
(To be clear, the C code also needs sanity tests, but
 it's way more reliable to test things on the Rust side)

---


rust/composefs/src/dumpfile.rs
```diff
@@ -445,4 +472,43 @@ mod tests {
...
+    #[test]
+    fn test_should_fail() {
+        const CASES: &[(&str, &str)] = &[
+            (
+                "long link",
+                include_str!("../../../tests/assets/should-fail-long-link.dump"),
+            ),
+            (
+                "no ftype",
+                include_str!("../../../tests/assets/should-fail-no-ftype.dump"),
```
cgwalters, line 496: Note I *didn't* need to change the Rust side for this one, it was already strictly validating the mode :trophy:

alexlarsson (approved):

PR #320, opened 2024-08-20
dump: Support `--filter`

cgwalters:
I'm looking at storing OCI container images with their metadata included in the image - giving a "single file" holding all the relevant bits for an image. Specifically a structure like this:

```
/manifest.json
/config.json
/rootfs
```

I want higher level tooling to be able to read the metadata efficiently. Of course we could directly parse the EROFS image...but a lot of pitfalls there. We could also mount the image but that's pretty inefficient for this.

Add support for filtering when generating a dump; this can operate fully unprivileged, and only requires some minor modifications to the library and CLI tooling.


alexlarsson (approved):
lgtm

PR #326, opened 2024-08-30
Replace custom clang-format bits with stock meson targets

cgwalters:
This is built into meson:
https://mesonbuild.com/Code-formatting.html


cgwalters:
Fun; I tried to update to ubuntu-24.04 across the board but it seems the fuse test starts erroring out with that...will defer that tech debt to later...

.github/workflows/test.yaml
```diff
@@ -6,6 +6,20 @@ permissions:
   actions: read
 
 jobs:
+  clang-format:
+    # To pick up a newer clang
+    runs-on: ubuntu-24.04
```
allisonkarlitskaya, line 11: Any reason not to move the main workflow over to 24.04?  We're probably gonna get exposed to that in the next couple of months anyway (wieh it becomes the new `ubuntu-latest`)... makes sense to shake out any issues ahead of time.

cgwalters: I added a comment about that https://github.com/containers/composefs/pull/326#issuecomment-2321626211

allisonkarlitskaya: Sorry.  Missed that.

allisonkarlitskaya (approved):

PR #329, opened 2024-09-04
Add 64bit rdev APIs

cgwalters:
I stumbled across the fact that on Linux/glibc at least `dev_t` is 64 bits, but we were truncating it to 32 bits.

Note that the type used in the `mkcomposefs` binary was already uint64_t but because C happily truncates/converts integers this was missed.

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -1009,12 +1009,24 @@ void lcfs_node_set_gid(struct lcfs_node_s *node, uint32_t gid)
 	node->inode.st_gid = gid;
 }
 
+// Deprecated: this truncates rdev to 32 bits
 uint32_t lcfs_node_get_rdev(struct lcfs_node_s *node)
+{
+	return (uint32_t)lcfs_node_get_rdev64(node);
+}
+
+uint64_t lcfs_node_get_rdev64(struct lcfs_node_s *node)
 {
 	return node->inode.st_rdev;
 }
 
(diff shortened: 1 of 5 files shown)


libcomposefs/lcfs-writer.c
```diff
@@ -1009,16 +1009,28 @@ void lcfs_node_set_gid(struct lcfs_node_s *node, uint32_t gid)
 	node->inode.st_gid = gid;
 }
 
+// Deprecated: this truncates rdev to 32 bits
```
allisonkarlitskaya, line 1012: `__attribute__((deprecated))`?

allisonkarlitskaya (commented):
I wonder if we could use a macro to direct users towards the new variant on recompile, dropping the old declaration from the headers (but keeping it defined to avoid breaking ABI).  Would save some churn...

cgwalters:
> I wonder if we could use a macro to direct users towards the new variant on recompile, dropping the old declaration from the headers (but keeping it defined to avoid breaking ABI). Would save some churn...

But if any callers had been interacting with this via `uint32_t` (as would be reasonable to do) and they *also* weren't using `-Wconversion` they'd not know that they need to take action. I think forcing the churn is the right thing here.

libcomposefs/lcfs-writer.c
cgwalters: Done and that also helped me notice that I forgot to port a consumer of the API here, thanks!

allisonkarlitskaya (approved):
lgtm

PR #330, opened 2024-09-04
rust: Accept 64 bit values for rdev

cgwalters:
Pairs with https://github.com/containers/composefs/pull/329 which adapts the C side to properly accept these too.

This breaks semver, we'll need to bump that.

### DIFF
--- a/rust/composefs/src/dumpfile.rs
+++ b/rust/composefs/src/dumpfile.rs
@@ -88,7 +88,7 @@ pub enum Item<'p> {
         /// Number of links
         nlink: u32,
         /// The device number
-        rdev: u32,
+        rdev: u64,
     },
     /// A symbolic link
     Symlink {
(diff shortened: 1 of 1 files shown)


jeckersb (approved):
:+1: 

It somehow never occurred to me that the fields on a struct-style enum variant are implicitly public.  There's [an RFC](https://github.com/rust-lang/rfcs/issues/3506) to allow them to be private though, and the discussion contains some tricks to achieve opaqueness using the current system.  Just thought I'd mention it here in case we'd really rather hide all of the internal details of `Item` and only expose it through accessor methods and/or via `Display`.

cgwalters:
Even if the struct data wasn't public presumably we would have made a public getter be `u32` though.

That said making the fields not be public we *could* make invalid states not be constructable though which is probably a good idea.

PR #333, opened 2024-09-06
libcomposefs: detect short erofs files

allisonkarlitskaya:
When attempting to read the header of the erofs file before mounting it, we verify that the read is successful, but not that the full header has been returned.  We then proceed to access the header, which means we could be reading uninitialized memory.

Add a check to verify that we've read the full header.  If not, return -EINVAL, which is what we already return in case the header was incorrect.


cgwalters (approved):
ooo :see_no_evil: Nice find!

I wonder if this one might also be related to https://github.com/containers/storage/issues/2042

cgwalters:
Hello @allisonkarlitskaya there's a request to relicense this repository, please comment on https://github.com/containers/composefs/issues/344

Issue #332, opened 2024-09-05
idea: "upper layer" (erofs) inside of initramfs


allisonkarlitskaya:
This is a really vague idea that I discussed with @cgwalters and @travier today.  They both said it belongs here as an issue.  At this point this is little more than a raw braindump.  There's a lot to think through and discuss.

The erofs produced by mkcomposefs on a reasonably complete `/usr` is on the order of double digits MB.  I've seen ~50MB generally, and it compresses well (down to more like 10MB).  The initramfs+kernel on my Silverblue system is low triple digits (~150MB, most of which is the initramfs).

It wouldn't be completely unreasonable, then, to have a complete static copy of the composefs "upper layer" erofs image *inside* of the UKI.  This would completely side-step quite a lot of thorny issues around binding the UKI to the correct deployment: all you'd need is the kernel image and the digest store.

How we get a UKI with this erofs inside of it could go two ways:

 - generate this on the end-user system by (deterministic magic) which lets us get a UKI which is bit-for-bit the same as the one we were expecting it to be.  We'd have some out-of-band signature somewhere (in some metadata that doesn't become part of the image) that we could then use for signing this.

 - push everything to the container image creation: the kernel image would be created as the last step of the image creation process.  This would involve running mkcomposefs inside of the container, on the contents of the container itself, and embedding the resulting blob into the UKI, which we'd then write to the container image at a well-known path.  Any signing that we might want to do as part of creating the image could happen at this point, inside of the image (or in another build stage and copied back into the final image).

The second approach has an extremely simple deployment strategy: just extract the container .tar directly into a composefs digest store (without creating the erofs).  The backing store should now contain all of the files that the erofs referred to.  Install the kernel image into the EFI ESP and you're done.

The second way seems wonderfully simple until you realize that there are some very serious drawbacks there:
 - we're essentially creating a new container format: the metadata about which files are part of the image is stored in the .tar of the image, but now also in the erofs that we put inside of the UKI.
 - which means, of course, that it's no longer possible to make casual modifications to the container to add a file or install an extra package or so: you need to regenerate the kernel image.  Maybe that's not so bad?

I think the second approach could be extremely nice for specific deployment scenarios, but it's a very different flavour than what has been promised for the "FROM fedora / ADD / RUN / ..." approach to OS customization.

So that takes us back to a reality where we probably want to support the first scenario of building the composefs and assembling the UKI on the end system.  That needs a *lot* of thinking...

[...]

PR #334, opened 2024-09-06
all: add 'copy' mount option

allisonkarlitskaya:
This copies the content of the named composefs image into a sealed memfd and mounts that, instead of using the underlying file.

The main benefit here is that the underlying filesystem isn't pinned by having one of its files used as the basis of the loopback device.  This is useful if you want to embed a composefs into an initramfs, for example.

There are also integrity benefits.  Without this option, the fsverity of the file is verified at mount time, but nothing stops the file from being modified after the fact.  With the copy option, we measure and compare the fsverity signature after copying into and sealing the memfd.

Right now this is implemented as a straight copy, but we may also add support for decompression at some point.  The erofs images compress reasonably well using general-purpose compression algorithms: a 52MB image of /usr on my system compresses down to 14MB with zstd and 13MB with xz.

### DIFF
--- a/libcomposefs/lcfs-mount.c
+++ b/libcomposefs/lcfs-mount.c
@@ -37,6 +37,8 @@
 
 #include <sys/syscall.h>
 #include <sys/mount.h>
+#include <sys/mman.h>
+
 #ifdef HAVE_FSCONFIG_CMD_CREATE_LINUX_MOUNT_H
 #include <linux/mount.h>
 #endif
@@ -666,6 +668,35 @@ static errint_t lcfs_mount(struct lcfs_mount_state_s *state)
 	return -EINVAL;
 }
 
+static int lcfs_mount_copy_and_seal(int fd) {
+	int memfd = memfd_create(CFS_MOUNT_SOURCE, MFD_CLOEXEC | MFD_ALLOW_SEALING);
+	if (memfd < 0)
+		return -errno;
+
+	// in the general case, copy_file_range() to a memfd doesn't work
+	ssize_t n;
+	do {
+		char buffer[1 << 20];  // 1MB
+		n = read(fd, buffer, sizeof buffer);
+		if (n < 0)
+			break;
+		n = write(memfd, buffer, n);
+	} while (n > 0);
+
+	if (n < 0) {
+		close(memfd);
+		return n;
+	}
+
+	if (fcntl(memfd, F_ADD_SEALS, F_SEAL_SEAL | F_SEAL_GROW | F_SEAL_SHRINK | F_SEAL_WRITE) < 0) {
+		int err = -errno;
+		close(memfd);
+		return err;
+	}
+
+	return memfd;
+}
+
 int lcfs_mount_fd(int fd, const char *mountpoint, struct lcfs_mount_options_s *options)
 {
 	struct lcfs_mount_state_s state = { .mountpoint = mountpoint,
@@ -680,7 +711,21 @@ int lcfs_mount_fd(int fd, const char *mountpoint, struct lcfs_mount_options_s *o
 		return -1;
 	}
 
+	int memfd = -1;
+	if (options->flags & LCFS_MOUNT_FLAGS_COPY) {
+		memfd = lcfs_mount_copy_and_seal(fd);
+		if (memfd < 0) {
+			return memfd;
+		} else {
+			state.fd = memfd;
+		}
+	}
+
 	res = lcfs_mount(&state);
+
+	if (memfd != -1)
+		close(memfd);
+
 	if (res < 0) {
 		errno = -res;
 		return -1;
@@ -708,10 +753,26 @@ int lcfs_mount_image(const char *path, const char *mountpoint,
 	if (fd < 0) {
 		return -1;
 	}
-	state.fd = fd;
+
+	int memfd = -1;
+	if (options->flags & LCFS_MOUNT_FLAGS_COPY) {
+		memfd = lcfs_mount_copy_and_seal(fd);
+		if (memfd < 0) {
+			errno = memfd;
+			return -1;
+		} else {
+			state.fd = memfd;
+		}
+	} else {
+		state.fd = fd;
+	}
 
 	res = lcfs_mount(&state);
+
 	close(fd);
+	if (memfd != -1)
+		close(memfd);
+
 	if (res < 0) {
 		errno = -res;
 		return -1;
--- a/tools/mountcomposefs.c
+++ b/tools/mountcomposefs.c
@@ -123,6 +123,7 @@ int main(int argc, char **argv)
 	bool opt_verity = false;
 	bool opt_tryverity = false;
 	bool opt_ro = false;
+	bool opt_copy = false;
 	int opt, fd, res, userns_fd;
 
 	while ((opt = getopt(argc, argv, "ht:o:")) != -1) {
@@ -198,6 +199,8 @@ int main(int argc, char **argv)
 			opt_ro = false;
 		} else if (strcmp("ro", key) == 0) {
 			opt_ro = true;
+		} else if (strcmp("copy", key) == 0) {
+			opt_copy = true;
 		} else {
 			errx(EXIT_FAILURE, "Unsupported option %s\n", key);
 		}
@@ -248,6 +251,8 @@ int main(int argc, char **argv)
 		options.flags |= LCFS_MOUNT_FLAGS_TRY_VERITY;
 	if (opt_ro)
 		options.flags |= LCFS_MOUNT_FLAGS_READONLY;
+	if (opt_copy)
+		options.flags |= LCFS_MOUNT_FLAGS_COPY;
 
 	if (opt_idmap != NULL) {
 		userns_fd = open(opt_idmap, O_RDONLY | O_CLOEXEC | O_NOCTTY);
--- a/libcomposefs/lcfs-mount.h
+++ b/libcomposefs/lcfs-mount.h
@@ -37,8 +37,9 @@ enum lcfs_mount_flags_t {
 	LCFS_MOUNT_FLAGS_READONLY = (1 << 1),
 	LCFS_MOUNT_FLAGS_IDMAP = (1 << 3),
 	LCFS_MOUNT_FLAGS_TRY_VERITY = (1 << 4),
+	LCFS_MOUNT_FLAGS_COPY = (1 << 5),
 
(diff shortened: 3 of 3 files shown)


allisonkarlitskaya:
This is not currently intended for merging, but might end up being part of the work we need to realize #332.

allisonkarlitskaya:
The verity stuff is wrong here, mostly because I didn't understand how it worked: it's enforced by the filesystem and we just have to verify that the fingerprint is the one we expected.  After we do that, we can't possibly read incorrect data.

Also: memfd doesn't support it, which means we'll end up with `ENOTTY` if we try that.

Instead, if we really want to inspect the data in the memfd, we should probably use the userspace `lcfs-fsverity` stuff to calculate the checksum for ourselves.  If we do that after we seal, we can be confident that nothing changes.  This provides us with a nice way to enforce a particular image, even if fsverity isn't enabled on the underlying filesystem.  Although: for that particular scenario, I wonder if it wouldn't be easier to just use the sha256sum...

I'm not sure if there's a real benefit to pursuing this, but in any case, if we do want to allow mixing `copy` and `digest` options, we need to retrieve the verity digest before we do the copy.

cgwalters:
> The verity stuff is wrong here, mostly because I didn't understand how it worked: it's enforced by the filesystem and we just have to verify that the fingerprint is the one we expected. After we do that, we can't possibly read incorrect data.

Right, we should be invoking `lcfs_validate_verity_fd` *before* copying (and then skip that verification on the memfd) I think.

cgwalters:
> The main benefit here is that the underlying filesystem isn't pinned by having one of its files used as the basis of the loopback device. This is useful if you want to embed a composefs into an initramfs, for example.

I think at least historically some initramfs code basically solved this with `rm -rf /*` after the pivot? But I guess it is perhaps cleaner to have a memfd instead of a tmpfs with one unlinked file.

travier:
I've not tested this change but I really like the approach. 👍🏻

tools/mountcomposefs.c
```diff
@@ -198,6 +199,8 @@ int main(int argc, char **argv)
 			opt_ro = false;
 		} else if (strcmp("ro", key) == 0) {
 			opt_ro = true;
+		} else if (strcmp("copy", key) == 0) {
```
alexlarsson, line 202: Need to add this to the manpage and the usage output.

libcomposefs/lcfs-mount.c
```diff
@@ -666,6 +668,35 @@ static errint_t lcfs_mount(struct lcfs_mount_state_s *state)
...
+static int lcfs_mount_copy_and_seal(int fd) {
+	int memfd = memfd_create(CFS_MOUNT_SOURCE, MFD_CLOEXEC | MFD_ALLOW_SEALING);
+	if (memfd < 0)
+		return -errno;
+
+	// in the general case, copy_file_range() to a memfd doesn't work
+	ssize_t n;
+	do {
+		char buffer[1 << 20];  // 1MB
+		n = read(fd, buffer, sizeof buffer);
```
alexlarsson, line 680: Need to handle EINTR.

libcomposefs/lcfs-mount.c
```diff
@@ -666,6 +668,35 @@ static errint_t lcfs_mount(struct lcfs_mount_state_s *state)
...
+		if (n < 0)
+			break;
+		n = write(memfd, buffer, n);
```
alexlarsson, line 683: Need to handle short writes, EINTR and write errors in general. I mean, i'm sure memfd writes don't get i/o errors, but you may be low on memory or whatever.

libcomposefs/lcfs-mount.c
```diff
@@ -666,6 +668,35 @@ static errint_t lcfs_mount(struct lcfs_mount_state_s *state)
 	return -EINVAL;
 }
 
+static int lcfs_mount_copy_and_seal(int fd) {
+	int memfd = memfd_create(CFS_MOUNT_SOURCE, MFD_CLOEXEC | MFD_ALLOW_SEALING);
```
alexlarsson, line 672: I'd use `cleanup_fd int memfd = -1;:` here and then `return steal_fd(&memfd)`; (Although i see we need to add a steal_fd for this.

libcomposefs/lcfs-mount.c
```diff
@@ -680,7 +711,21 @@ int lcfs_mount_fd(int fd, const char *mountpoint, struct lcfs_mount_options_s *o
 		return -1;
 	}
 
+	int memfd = -1;
```
alexlarsson, line 714: cleanup_fd

libcomposefs/lcfs-mount.c
```diff
@@ -680,7 +711,21 @@ int lcfs_mount_fd(int fd, const char *mountpoint, struct lcfs_mount_options_s *o
...
+	if (options->flags & LCFS_MOUNT_FLAGS_COPY) {
+		memfd = lcfs_mount_copy_and_seal(fd);
+		if (memfd < 0) {
+			return memfd;
+		} else {
```
alexlarsson, line 719: I'd rather skip the else clause in this kind of "didn't get error" codepath. I.e.:
```
if (memfd < 0) {
  return memfd:
}
state.fd = memfd;
```

alexlarsson:
So, yeah, the integrity parts here are not correct, but I think this makes sense in some cases to eg. allow the source fs to be unmounted, but also as a general robustness thing. Like, its nice that we can guarantee "right data or EIO", but with this, we can guarantee "right data" (at least for the fs metadata), at the cost of slower mount and more memory use.

alexlarsson:
I'm not sure how much this matters, but we can also with this get fs-verity validation if the file on disk doesn't have fs-verity (i.e. if it is on FAT or whatever), because we can manually valdiate the digest when copying into the memfd. Of course, we would still need fs-verity for the backing files.

Again, not sure if this is practically useful, but worth considering.

allisonkarlitskaya:
> I'm not sure how much this matters, but we can also with this get fs-verity validation if the file on disk doesn't have fs-verity (i.e. if it is on FAT or whatever), because we can manually valdiate the digest when copying into the memfd. Of course, we would still need fs-verity for the backing files.
> 
> Again, not sure if this is practically useful, but worth considering.

It's absolutely my plan to do the fsverity checking in userspace on the already-sealed memfd.

allisonkarlitskaya:
> It's absolutely my plan to do the fsverity checking in userspace on the already-sealed memfd.

...until we find some kernel hacker who is willing to add support for fsverity to tmpfs/memfd...

alexlarsson:
> It's absolutely my plan to do the fsverity checking in userspace on the already-sealed memfd.

I don't think this is necessary if we validated the digest on the fd after we opened it. We just copied data that was already validated (by the read call). How could it have become modified after that? 

I mean, in theory it could somehow become modified in memory after reading by some ptrace attacker, but that is also true for userspace validation after the fact (i.e. the code that checks the fsverity digest of the data read from the sealed fd could have been modified). Really, the security model is that mount.composefs comes from a signed trusted initrd, and runs only code that we trust, so there is no "attacker".

alexlarsson:
Btw, I would do the fsverity manual computation in the copy-to-memfd loop, then you better parallelize the "read next block" and the "compute digest of block".

allisonkarlitskaya:
I think someone could probably also open("/proc/($pidof mount.composefs)/fd/3") or whatever the memfd is and modify it like that.  I'm not super-concerned about that case.  Supporting filesystems that lack their own fsverity is more what I'm aiming for here, and I agree that this could be done while streaming the file from the disk...

alexlarsson:
> I think someone could probably also open("/proc/($pidof mount.composefs)/fd/3") or whatever the memfd is and modify it like that. I'm not super-concerned about that case. Supporting filesystems that lack their own fsverity is more what I'm aiming for here, and I agree that this could be done while streaming the file from the disk...

Not if it is sealed

alexlarsson:
That said, you will still definitely need fs-verity for the filesystem backing the file objects. No way out of that.

allisonkarlitskaya:
> That said, you will still definitely need fs-verity for the filesystem backing the file objects. No way out of that.

This much is clear.  What we're imagining, though, is that the erofs layer lives in the initramfs, where it doesn't get fsverity.  The digest store is on a "real" filesystem, though.

alexlarsson:
> This much is clear. What we're imagining, though, is that the erofs layer lives in the initramfs, where it doesn't get fsverity. The digest store is on a "real" filesystem, though.

This sounds very interesting. However, in that case the file is typically trusted by being on the initrd itself, which is signed, so fs-verity of it is not strictly necessary then. (In particular, because generally all fs-verity digests we verify against are generally trustworthy due to having the public key in the initrd.)

cgwalters:
Yes agreed, there is no need for us to enforce fsverity of the erofs in this model.

Maybe we change `lcfs_validate_verity_fd()` to check if the source fd is a memfd, and if it is then we require that it's sealed?

---

Actually though...will Linux ever swap out a memfd contents? If it can, then we *do* need to enforce verity for it right?

cgwalters:
Digging a bit there's nowadays `memfd_secret()` which is defined to disallow swapping...we could use that. Although, it seems possible the "The memory region is removed from the kernel page tables" part would break us loopback mounting it?

cgwalters:
Of course, the other alternative design to think about here is basically: "don't put an EROFS in the initramfs, just put it unpacked in e.g. `/rootfs` in the initramfs". No loopback mounts, memfd etc. needed. We'd just have something like `LCFS_MOUNT_DIR` or so that tells the mounting code it's a plain old directory file descriptor.

However, the swapping thing still seems like a concern to me...actually look, there's already a `noswap` option for tmpfs, so we'd just need to be sure that the initramfs tmpfs is set up with that option.

libcomposefs/lcfs-mount.c
cgwalters: We already have `copy_file_data_classic()`

alexlarsson:
> Of course, the other alternative design to think about here is basically: "don't put an EROFS in the initramfs, just put it unpacked in e.g. `/rootfs` in the initramfs". No loopback mounts, memfd etc. needed. We'd just have something like `LCFS_MOUNT_DIR` or so that tells the mounting code it's a plain old directory file descriptor.

Does this work? For example, does the initramfs tar unpacking support full xattrs? Also, my guess is that a full in-memory tmpfs of the rootfs metadata will use a lot more ram than just the erofs image.

allisonkarlitskaya:
I think we've decided not to do this.

PR #337, opened 2024-09-09
A few cleanups around symlinks

cgwalters:
mkcomposefs: Drop duplicate invocation of lcfs_node_set_mode()

Just noticed this while I was reading the code for other reasons.

---

Add fallible lcfs_node_try_set_mode()

Previously I added code to validate this but it happened much
later. The mode (type) of a file is a very fundamental property
and we should constrain its state as early as possible.

---

Add lcfs_node_set_symlink_payload

Historically we accept any arbitrary data in the dumpfile
for the file size for symlink, and end up ignoring it
ultimately when we write the EROFS. However previously
it was pretty confusing as the in-memory node data
could have a bogus size.

Add an API and use it both in the mkcomposefs path *and*
in the "reread from EROFS path" to perform more consistent
validation for symlink targets. This just wraps the
previous `lcfs_node_set_payload()` API, inheriting its
checks for length.

---

### DIFF
--- a/libcomposefs/lcfs-writer-erofs.c
+++ b/libcomposefs/lcfs-writer-erofs.c
@@ -1746,7 +1746,7 @@ static struct lcfs_node_s *lcfs_build_node_from_image(struct lcfs_image_data *da
 
 		memcpy(name_buf, tail_data, file_size);
 		name_buf[file_size] = 0;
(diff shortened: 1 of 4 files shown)


jeckersb (approved):
LGTM

PR #338, opened 2024-09-09
tests: Move sha verification after other checks

cgwalters:
When testing out changes to a dumpfile, today the sha check will always fail first if I don't go and change the expected value, but most of the time what I want to know is whether or not it passes the *other* tests at least.

Move the checksum verification last, and also while we're here print an "ok" value.


jeckersb (approved):
:+1:

Issue #327, opened 2024-09-04
`mkcomposefs --from-file` ignores last line without newline


cgwalters:
It took me an unfortunate amount of time to figure out that there's a bug in `tree_from_dump()` where basically we ignore the last chunk of a dumpfile if it doesn't end in a newline.

```
$ cat dump-nonl.txt
/ 4096 40555 222 0 0 10 0.0 - - - trusted.foo1=bar-1 user.foo2=bar-2
/a 0 100644 1 0 0 0 0.0 - - -
/b 0 100644 1 0 0 0 0.0 - - -
$ open /tmp/dump-nonl.txt | base64 -w 0
LyA0MDk2IDQwNTU1IDIyMiAwIDAgMTAgMC4wIC0gLSAtIHRydXN0ZWQuZm9vMT1iYXItMSB1c2VyLmZvbzI9YmFyLTIKL2EgMCAxMDA2NDQgMSAwIDAgMCAwLjAgLSAtIC0KL2IgMCAxMDA2NDQgMSAwIDAgMCAwLjAgLSAtIC0=
$ mkcomposefs --from-file dump-nonl.txt out.cfs
$ composefs-info dump out.cfs
/ 4096 40555 2 0 0 0 0.0 - - - trusted.foo1=bar-1 user.foo2=bar-2
/a 0 100644 1 0 0 0 0.0 - - -
$
```

The question is...what do we do about it? I guess it's *probably* unlikely someone was relying on this behavior, but it's hard to say that definitively.

I guess I lean towards changing the behavior to be an error (i.e. a trailing newline is required). But I am curious if others have opinions.

jeckersb:
I lean towards allowing the no-newline case, only because `composefs-dump(5)` says:
```
The file format is very simple, with one file per line[...]

Fields are separated by a single space, and lines by a single newline.
```
Which as-documented to me only requires them to be separated by a newline but not terminated by one.  So I think to keep the previously-published behavior it would be better to allow it.  If it's disallowed, then at the very least the docs should be updated to reflect the requirement.

cgwalters:
OK I've updated https://github.com/containers/composefs/pull/331

PR #340, opened 2024-09-10
mkcomposefs: Add a hidden CFS_PARSE_STRICT

cgwalters:
Right now in some cases we have "loose" semantics; a simple case I noticed earlier is we accepted whatever input size for a symlink but ignored it.

I also noticed we accepted (and ignored) a nonzero rdev for a non-device.

We have prior cases around missing trailing newline, etc.

In preparation for a more official "mkcomposefs --strict" mode, add a hidden environment variable and use it to validate some of the above.

There's other bits missing though - for example in strict mode I'd like to hard require that the xattrs are already in canonical form (sorted, no duplicates).

This touches on a general topic that there's "strict" and then something even stronger around "normal form" where we disallow any ambiguous input such as duplicate `//` in filenames etc.

That'd be a good bit more work. For now let's just lay the groundwork for stricter validation.

### DIFF
--- a/tests/test-checksums.sh
+++ b/tests/test-checksums.sh
@@ -14,6 +14,12 @@ tmpfile=$(mktemp --tmpdir lcfs-test.XXXXXX)
 tmpfile2=$(mktemp --tmpdir lcfs-test.XXXXXX)
 trap 'rm -rf -- "$tmpfile" "$tmpfile2"' EXIT
 
+# Ensure our builtin dumpfiles are strict, minus some exceptions
+declare -A nonstrict
+nonstrict=(
+    ["no-newline.dump"]="1"
+)
+
 for format in erofs ; do
     for file in ${TEST_ASSETS} ; do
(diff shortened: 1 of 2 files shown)


giuseppe (approved):
would it make sense to always have these checks instead of a "strict" mode?

Either way, I am fine with it, LGTM

cgwalters:
> would it make sense to always have these checks instead of a "strict" mode?

The thing that has me hesitant is we know (from PRs and issues here) that there are a set of composefs users who are working today, and I want to avoid breaking them.

See e.g. https://github.com/containers/composefs/issues/327#issuecomment-2332603274 where John (correctly) argued to accept the no-newline case - which I explicitly switched back to denying in strict mode here.

We can ask composefs users to try out "strict mode" in the future and maybe then consider turning it on by default...or probably before that, just emitting warnings in non-strict mode if we find strict mode violations, etc.

PR #341, opened 2024-09-11
lib: Move validation out of lcfs_node_add_child()

cgwalters:
Unfortunately at least ostree does:

node = lcfs_node_new()
lcfs_node_add_child(parent, node);
lcfs_node_set_mode(node, ...)
So we basically have a completely uninitialized node as part of the tree.

This failed our basic "validate the mode" logic.

We can't do any validation at all in `lcfs_node_add_child()` so move it to the first time we walk the tree as part of `lcfs_node_write_to()`.

Also as part of fixing this: Today we don't really have coverage of the C library as it may be used by external consumers. Add a unit test to verify this, and we can build on this more.

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -390,12 +390,6 @@ int lcfs_write_to(struct lcfs_node_s *root, struct lcfs_write_options_s *options
 	struct lcfs_ctx_s *ctx;
 	int res;
 
-	// Verify the root; because lcfs_node_add_child also verifies children,
-	// we should have sanity checked all nodes.
-	if (lcfs_node_last_ditch_validation(root) < 0) {
-		return -1;
-	}
-
 	/* Check for unknown flags */
 	if ((options->flags & ~LCFS_FLAGS_MASK) != 0) {
 		errno = EINVAL;
@@ -1174,7 +1168,7 @@ struct lcfs_node_s *lcfs_node_get_hardlink_target(struct lcfs_node_s *node)
 // invalid states, but don't return errors; this will try
 // to perform validation when the node is passed to a function
 // that does return an error.
-int lcfs_node_last_ditch_validation(struct lcfs_node_s *node)
+int lcfs_node_validate(struct lcfs_node_s *node)
 {
 	// Hardlinks should have mode 0
 	if (node->link_to == NULL) {
@@ -1189,10 +1183,6 @@ int lcfs_node_add_child(struct lcfs_node_s *parent, struct lcfs_node_s *child,
 {
 	struct lcfs_node_s **new_children;
 	size_t new_capacity;
-
-	if (lcfs_node_last_ditch_validation(child) < 0) {
-		return -1;
-	}
 
 	if ((parent->inode.st_mode & S_IFMT) != S_IFDIR) {
 		errno = ENOTDIR;
--- a/tests/test-lcfs.c
+++ b/tests/test-lcfs.c
@@ -1,0 +1,81 @@
+#define _GNU_SOURCE
+
+#include "lcfs-writer.h"
+#include <assert.h>
+#include <errno.h>
+
+static inline void lcfs_node_unrefp(struct lcfs_node_s **nodep)
+{
+	if (*nodep != NULL) {
+		lcfs_node_unref(*nodep);
+		*nodep = NULL;
+	}
+}
+#define cleanup_node __attribute__((cleanup(lcfs_node_unrefp)))
+
+static ssize_t write_cb(void *_file, void *buf, size_t count)
+{
+	FILE *file = _file;
+
+	return fwrite(buf, 1, count, file);
+}
+
+static int testwrite_node(struct lcfs_node_s *node)
+{
+	char *bufp = NULL;
+	size_t bufsz = 0;
+	FILE *buf = open_memstream(&bufp, &bufsz);
+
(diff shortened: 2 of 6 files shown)


cgwalters:
Ah yeah of course since this is C a ~5 line unit test has a double-unref, thankfully caught by ASAN.

allisonkarlitskaya (commented):
The patch looks fine as it is but it's sort of clear that the API use by ostree is not what was intended here...

Would it be super disruptive to assert that it's invalid to male changes to the node after adding it to the parent?  I guess that would more or less be an ABI break and need an soname bump, right?  I'm thinking about the recent `_rdev64()` changes...  What's our guarantee here?

cgwalters:
Yeah, agree. Here's one thing I've been thinking about as followups to all this:

- Add `lcfs_node_new_symlink()` `lcfs_node_new_inline_file()` `lcfs_node_new_dev()` etc that take their *required* properties as constructors (very much like how things are modeled on the Rust side)
- These APIs also set a `node->strict = true` property that enforces more validation
- When `lcfs_node_add_child() is invoked on a node with strict = true, any further mutation APIs return an error?

Actually we could just add `lcfs_node_new_dev()` now and *not* ship `lcfs_node_set_rdev64` I guess.

>  guess that would more or less be an ABI break and need an soname bump, right?

Right, I don't want to scope that into this at this time. It would be painful.

allisonkarlitskaya:
> Add `lcfs_node_new_symlink()` `lcfs_node_new_inline_file()` `lcfs_node_new_dev()` etc that take their _required_ properties as constructors (very much like how things are modeled on the Rust side)

This sounds nice.

Is your plan to land this PR, add those APIs, then back out the rdev64 setter, then deprecate the existing `_new()` API as a way to more or less remove the ability to create `->strict == false` nodes?  And then port ostree over?  Because it's sort of at cross purposes to this PR to move validation to 'later' because of node mutability if we intend to head in a direction where the nodes are no longer mutable...

cgwalters:
My hope was to land this PR basically as is alongside https://github.com/containers/composefs/pull/342 and get a new release out and leave other things to followup.

cgwalters:
Adding a new API is going to need more tests, and we'd need to rework the fuzzing entrypoint to support *both* the old and new API, which is quite doable but also not trivial, so I'd rather get fixes out for what we have now as a release.

libcomposefs/lcfs-writer.c
```diff
@@ -390,12 +390,6 @@ int lcfs_write_to(struct lcfs_node_s *root, struct lcfs_write_options_s *options
 	struct lcfs_ctx_s *ctx;
 	int res;
 
-	// Verify the root; because lcfs_node_add_child also verifies children,
-	// we should have sanity checked all nodes.
-	if (lcfs_node_last_ditch_validation(root) < 0) {
```
allisonkarlitskaya, line 395: One very minor nag: we handle validation of the node from the main writer code currently but your patch moves it to the erofs implementation code.  Of course, there *are* no other implementations... but it seems a bit odd.

I guess the implementation is the thing iterating the nodes in the end, though, so there's not really much of another way...

tests/test-lcfs.c
```diff
@@ -0,0 +1,32 @@
...
+	lcfs_node_set_mode(node, S_IFDIR | 0755);
+	// libostree today does this pattern of creating an empty (uninitialized)
+	// child and passing it to lcfs_node_add_child first. Verify this
+	// continues to work for the forseeable future.
+	cleanup_node struct lcfs_node_s *child = lcfs_node_new();
+	int r = lcfs_node_add_child(node, child, "somechild");
+	assert(r == 0);
+	// Adding child took ownership
+	child = NULL;
+}
```
allisonkarlitskaya, line 27: Can you try writing an erofs here and making sure the validation error is flagged at that point?

cgwalters: Sure done

tests/test-lcfs.c
```diff
@@ -0,0 +1,81 @@
...
+	FILE *file = _file;
+
+	return fwrite(buf, 1, count, file);
+}
+
+static int testwrite_node(struct lcfs_node_s *node)
+{
+	char *bufp = NULL;
+	size_t bufsz = 0;
+	FILE *buf = open_memstream(&bufp, &bufsz);
```
allisonkarlitskaya, line 27: The memstream was more than I was expecting!

tests/test-lcfs.c
```diff
@@ -0,0 +1,81 @@
...
+
+	struct lcfs_write_options_s options = { 0 };
+	options.format = LCFS_FORMAT_EROFS;
+	options.version = 1;
+	options.max_version = 1;
+	options.file = buf;
+	options.file_write_cb = write_cb;
+
+	int r = lcfs_write_to(node, &options);
+	int saved_errno = errno;
```
allisonkarlitskaya, line 37: It took me a real long time to convince myself that this was right, but indeed we set `errno = EINVAL` in the mode-validation code.  Is there some story behind the various approaches to error handling in libcomposefs?  The mount code seems to prefer returning negative errnos, at least internally...

allisonkarlitskaya (approved):
Thanks for the test!  This looks good to me.

tests/test-lcfs.c
cgwalters: > The mount code seems to prefer returning negative errnos, at least internally...

It's quite possible that I or someone else messed up the sign at some point :cry:

PR #342, opened 2024-09-12
Results from running honggfuzz, plus some more manual auditing

cgwalters:
A while ago while working on the Rust composefs bindings I found [a bug](https://github.com/containers/composefs/pull/310/commits/b06ea9b47315057a6daf6e2ffbab898358415fe3) in the dumpfile parser where the process would crash if provided an unexpected value.

Today for the Rust bindings and the [containers/image library](https://github.com/containers/image) because we're forking off a separate process (that we can apply isolation to and low privilege overall) even an assertion failure is relatively benign. And such tools usually wouldn't produce a dumpfile malformed in that way. But still, it caused me to question whether there were *other* bugs.

For codebases that are linking in the libcomposefs library, process aborts are obviously more problematic.

This series is the result of running [honggfuzz](https://honggfuzz.dev/) and quite a lot of time spent learning this codebase.

Thanks to `@hsiangkao @alexlarsson @jeckersb @giuseppe @allisonkarlitskaya` for their assistance with reviews and discussion of this!

# Composefs and security

I don't think there's cause to classify this as a CVE yet, as we didn't really formally declare that `mkcomposefs --from-file` is a "security boundary". At the current time, if you are using composefs in a context where it is being provided potentially malicious input, it is strongly recommended to ensure that the input is validated and sanitized as much as possible, alongside use of hardening and fuzzing techniques. We have developed [Rust bindings](https://crates.io/crates/composefs) for example which do some of this, and work will continue on that.

This said, we will likely aim to make "mkcomposefs --from-file" a security boundary in the future (e.g. guaranteeing that it will always either return an error or generate a valid EROFS) and allocate CVEs for any further flaws.

# Composefs and reliability

composefs has been used reliably in many cases for a while now; in general almost all of these issues could only occur for extreme boundary conditions such as:

- symlinks close to the PATH_MAX (4096 byte) limit
- extremely large extended attributes
- Files claiming to be *extremely* large (> 5 EiB)

Hence, I do not think there is a need for existing composefs users to consider this a critical fix.

Most of these cases also again would fail with an assertion - at most a denial of service. However in *some* cases we would produce a corrupted EROFS blob - and in the cases I tested, the Linux kernel EROFS parser successfully detected the corruption.

The EROFS kernel code is actively fuzzed and fixed when bugs are found (e.g. [this recent one](https://lists.ozlabs.org/pipermail/linux-erofs/2024-September/011750.html)) but still a goal for this project is to avoid exposing the Linux kernel code to a potentially corrupted EROFS at all.

(A core part of the idea of composefs is that the EROFS metadata doesn't need to be transported "on the wire", it can be re-synthesized reproducibly from other metadata, such as existing OCI tar layers, or ostree commits, etc.)

---

mkcomposefs: Reject missing = in xattrs

The man page was clear that xattrs required a `=`. The code
was loose and took this case to mean an xattr with an empty
value.

Change to hard rejecting this, as it's *probably* a mistake
by anyone doing this.

This also makes silly mistakes in hand-crafting dumpfiles
a *lot* more obvious, as previously an extra `-` for example
was interpreted as an xattr with the name `-`.

---

mkcomposefs: Reject excessive xattrs early

Because we call `lcfs_node_set_xattr()` a loop
which searches all previous values for duplicates, we
get O(N^2) behavior. Cap the maximum size of the input line.

---

rust/parser: Reject too many xattrs

This isn't fully strict, but it caps obviously broken input. Stricter
validation will happen on the C side for now.

---

writer: Disallow self-referential hardlinks

This one currently results in a stack overflow.

Credit: https://github.com/AFLplusplus/AFLplusplus

---

fuzzing: Also write erofs, and re-read it

It took me an embarrassing amount of time to realize
why the existing fuzzing code wasn't finding the same
things I was seeing just fuzzing "externally" with
AFL; it was because we were only fuzzing the parser,
not actually writing the EROFS.

Add that to the dedicated fuzzing path and while
we're here also re-read it back to exercise even
more code.

---

rust/dumpfile: Deny embedded NUL byte in paths and xattr names

There is no reason to support this, so don't. It turns out the
C side was loose on this, but we should be strict.

---

mkcomposefs: Reject embedded NUL characters in dumpfile

At least honggfuzz very quickly discovered that split_line() sloppily allows
embedded NUL characters, and its generated dumpfiles contain a lot of them
and make them unreadable by default.

We didn't document support for embedded NULs, and it only introduces
ambiguity in parsing, so let's just reject this early on.

---

writer: Work around whiteout chardevs with nonzero size

Even though it makes no sense for character devices to have
a nonzero size, today we just faithfully copy that into the
erofs by default.

Now, a corner case rears its head when that device appears
to be a whiteout, because we end up converting it to a regular
file with a special "escaped whiteout" xattr.

When we do that, ensure we also canonialize its size to zero
because that's what the re-reading path expects.

It would honestly probably be better to just force all
character devices to have zero size, but this is a targeted fix
to clear this fuzzing case off the board.

---

writer: Cap maximum size of all xattr data

I finally dug into the EROFS on-disk format a bit and this attempts
to a bit more accurately track input xattr sizes and deny additions
which would overflow EROFS' core limit of UINT16_MAX.

However...we have a further problem in that we add our *own*
xattrs to implement things like the overlayfs redirects, so let's
cap "external" xattrs to only UINT16_MAX / 2 - I think this
is still quite a lot. Add `#define`s for these so external
code can also query the limits.

Changed in v1:
- Account for worst case alignment by assuming 3 extra bytes per
  xattr

---

writer: Ensure re-setting xattrs are re-validated

We did a lot of work to limit the size of xattrs, but
honggfuzz quickly figured out it was possible to bypass
this by specifying an xattr multiple times; the second
copy had special "replace existing value" code that
didn't check limits.

Fix this by always calling the "remove xattr" API when
setting (but ignoring the value). This ensures we
are consistently tracking limits.

---

Handle long symlinks with xattrs

Credit: https://github.com/google/honggfuzz

With a long symlink including xattrs, we end up needing
to spill the symlink content into a separate block.

---

writer: Reject too large files

We don't currently handle non-inline files whose chunking would exceed
a single block. That limit is currently around 8 PiB which I am fairly
confident no one is going to try to put inside a composefs in
the near future.

We could lift this restriction in the future by writing multiple blocks,
but for now this fixes a crash detected by honggfuzz.

---

tests: Add a long symlink case

This exercises "symlink almost at maximum".

---

writer: Add a few more assertions

On general principle.

---

erofs: Reject obviously bogus non-inline file sizes

This avoids us trying to malloc() something ridiculous.

---

mkcomposefs: Don't set content on symlinks

Symlinks cannot have content. honggfuzz figured out that a
symlink with an "empty content" ended up resetting `st_size`
on the inode, confusing later code.

Error out if we have non-empty content on symlinks, and under
no conditions invoke `lcfs_node_set_content()` on a symlink.

---

erofs: Use sizeof(xattr struct) and not 4

Not hardcoding 4 here makes it more obvious which structure is
being referenced.

---

erofs: Add an assertion for xattr sizes

I was looking at whether this path is buggy; ensure that
the size we computed for xattrs in the inode path matches
up with what we end up writing to disk.

---

Handle non-inline files that need to be EROFS_INODE_CHUNK_BASED

---

Add limits and recommendations for inline content

The role of composefs is to be metadata, delegating most "regular file"
content to a backing store.

However we do support inline files (the simple case is a *zero sized*
file which happens surprisingly often); and we had an internal heuristic
of inlining files that are 64 bytes or smaller, which is less
than the SHA-256 checksum we'd need to follow them to the real
filesystem.

However...dumpfiles and `lcfs_node_set_content()` and our internal
infrastructure make no attempt to limit the size of inline files
at all.

One problem I realized here while working on fuzzing is that
there's a strict dependency between the size on the
dumpline and the inline content but many fuzzers are bad at
detecting these "strict data dependencies". So we also
change the code (only when compiled in fuzzing mode) to ignore
whatever the size is and accept what we find in the inline content.

This helped honggfuzz quickly find that we didn't handle a lot
of inline file content size at all.

---


cgwalters:
OK yeah...I wasn't running honggfuzz or even my local builds with ASAN+UBSAN enabled, very clearly should have been doing that. But in fairness to me the thing ASAN found was an existing memory leak, not a new one.

cgwalters:
Fixed textual conflict with https://github.com/containers/composefs/pull/340

jeckersb (commented):
Huge :+1: from me, this is great stuff.

Since composefs is pretty new to me, I'd like to see a sanity check from @alexlarsson or @giuseppe before we merge this, since I'm likely to overlook something and I trust them way more than myself :smile:

cgwalters:
The ones that haven't been reviewed in depth are probably:

- https://github.com/containers/composefs/pull/342/commits/2e44f5816b5d4e5099f8ff4074c9f93f3cf99e35
- https://github.com/containers/composefs/pull/342/commits/3ff7c6f608b2780bebeca59716598919c45fcd01

Also https://github.com/containers/composefs/pull/342/commits/6e12303c66e3c7601e985109538e2158d577ddc3 is a nontrivial one but I'm more confident in it.

giuseppe (approved):
Screenshot from 2024-09-13 08-31-36

LGTM

libcomposefs/lcfs-writer-erofs.c
```diff
@@ -500,6 +498,18 @@ static uint32_t compute_erofs_chunk_bitsize(struct lcfs_node_s *node)
 	return chunkbits;
 }
 
+/* In composefs for non-inline files we just have a "stub" block filled
+ * with a small amount of data, because the real data lives in the underlying
+ * filesystem. The chunks are filled with 0xFF.
+ */
```
alexlarsson, line 504: This comment isn't quite right. what happens is that we want to make the file look like it has the right size, but filled with zeros. But we want to do this in a "hole" way, i.e. without storing a bunch of zeros. In the chunked format the file data is a list of offsets to blocks of a specific size (the chunk format). We chose the maximum size for the chunk format that is smaller than the file we need (so typically we'll only use one chunk). But then, instead of using a real pointer we use the EROFS_NULL_ADDR which erofs uses for holes.

alexlarsson:
These all looks good to me, sans the comment above. My main worry with these changes are that some actually change the output of the erofs image, which would break backwards compat due to changing the digest. However, as far as I could see the changes would only happen when something previously would have generated a broken erofs image, so I don't think it is a problem really.

libcomposefs/lcfs-writer-erofs.c
alexlarsson: (ie. EROFS_NULL_ADDR makes erofs produce a bunch of zeros for that block)

cgwalters: Thanks! I'll fix that one up in a followup.

cgwalters:
> My main worry with these changes are that some actually change the output of the erofs image, which would break backwards compat due to changing the digest.

Right, quite often when hacking on things here and experimenting I did accidentally change output, but yes the CI tests cover this well.

PR #345, opened 2024-09-13
writer: Clarifications around our usage of EROFS_NULL_ADDR

cgwalters:
- Followup from review comments to clarify how this works
- Rename the function to further reinforce it's not "empty" it's actually the special null addr
- Actually use the `EROFS_NULL_ADDR` constant when we do the writing to make the connection more obvious
- When reading back an erofs, at least verify that it's chunk based as extra validation.

### DIFF
--- a/libcomposefs/lcfs-writer-erofs.c
+++ b/libcomposefs/lcfs-writer-erofs.c
@@ -498,10 +498,9 @@ static uint32_t erofs_compute_chunk_bitsize(uint64_t file_size)
 	return chunkbits;
 }
 
-/* In composefs for non-inline files we just have a "stub" block filled
- * with a small amount of data, because the real data lives in the underlying
- * filesystem. The chunks are filled with 0xFF.
- */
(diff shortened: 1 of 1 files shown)


libcomposefs/lcfs-writer-erofs.c
```diff
@@ -818,10 +817,12 @@ static int write_erofs_dentries(struct lcfs_ctx_s *ctx, struct lcfs_node_s *node
 	return 0;
 }
 
-static int write_empty_chunks(struct lcfs_ctx_s *ctx, uint32_t chunk_count)
+// Each chunk pointer appears as if it's filled with zeros using the dedicated
+// EROFS_NULL_ADDR.
+static int write_nullptr_chunks(struct lcfs_ctx_s *ctx, uint32_t chunk_count)
```
hsiangkao, line 822: although we usually call it as holes ;)

hsiangkao (approved):
LGTM

alexlarsson (approved):

PR #346, opened 2024-09-13
Release 1.0.5

cgwalters:
The main feature in this release is we ran honggfuzz and fixed a *lot* of bugs from that. These were all boundary conditions (extremely long symlinks, enormous files) and I am not aware of anyone hitting them in practice.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project(
   'composefs',
   'c',
-  version : '1.0.5',
+  version : '1.0.6',
   default_options : [
     'c_std=c99',
     'warning_level=1',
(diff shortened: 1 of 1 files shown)


alexlarsson:
Maybe we can finish the relicensing before the release?

cgwalters:
That seems a bit unlikely to happen fast, there's probably 10 other people that would need to agree from what I can see.

That'd be a good feature for the next release - it'd be good to ensure we have a regular cadence.

cgwalters:
Well, git log will forever record this as "Release 1.0.5" but obviously the title should have been "Release 1.0.6"

PR #353, opened 2024-09-20
rust: Another attempt to fix docs.rs build

cgwalters:
The build on docs.rs is still failing due to missing the C library. The standard thing to do here is just not to look for it when building docs.

This time I copied the setup from gtk4-rs; the newer pattern here seems to be directly using `rustc-args` and `rustdoc-args` instead of exposing a `dox` feature.

What I believe I missed in the last attempt was the copy of `package.metadata.docs.rs` in the `composefs`
crate.


jeckersb (approved):
:crossed_fingers:

cgwalters:
[success](https://docs.rs/composefs/latest/composefs/) :tada:

PR #355, opened 2024-09-20
mkcomposefs: make it cpuset aware

giuseppe:
A tiny improvement that I've noticed while looking at the code.

use sched_getaffinity to get the CPUs that can be used.

### DIFF
--- a/tools/mkcomposefs.c
+++ b/tools/mkcomposefs.c
@@ -38,6 +38,7 @@
 #include <linux/fsverity.h>
 #include <linux/fs.h>
 #include <pthread.h>
+#include <sched.h>
 #include <sys/sysinfo.h>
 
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Looks sane to me. One thing I just did was a quick [code search](https://github.com/search?q=lang%3Ac+sched_getaffinity+-is%3Afork&type=code&p=2) and yeah that led to [the exact same code in swift](https://github.com/search?q=lang%3Ac+sched_getaffinity+-is%3Afork&type=code&p=2).

That said I think an even nicer thing for us to consider in the future is support for the [GNU make jobserver protocol](https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html) which gives us a way to integrate with other build systems.

PR #357, opened 2024-09-23
Expose optimized fsverity interface to Rust

cgwalters:
On the C side we have two functions:

- `lcfs_compute_fsverity_from_fd`
- `lfs_fd_get_fsverity`

The first is currently defined to operate purely in memory. The second checks if fsverity is already enabled on the fd, and if so just asks the kernel for it; falling back to the first case otherwise.

I'm actually not entirely sure if we really need both; it seems like if we have a fd we'd basically always want to use the kernel code if possible.

Out of conservatism for now, expose both on the composefs-sys Rust binding side. But on the high level interface just change what we're using to be backed by `lcfs_fd_get_fsverity()`.

If someone really cares later of course we can add sugar for the pure compute path in the high level Rust API too.

### DIFF
--- a/rust/composefs-sys/src/lib.rs
+++ b/rust/composefs-sys/src/lib.rs
@@ -8,6 +8,7 @@ extern "C" {
         digest: *mut u8,
         fd: std::os::raw::c_int,
     ) -> std::os::raw::c_int;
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
I kinda wonder if we want to rather implement more of this stuff in Rust.....

giuseppe (approved):
LGTM

Issue #364, opened 2024-09-26
Build failure with Clang 18


ziyao233:
Building with clang 18, with following steps on master,

```
meson setup . build
meson compile -C build
```

results in error like

```
../libcomposefs/lcfs-fsverity.h:20:32: warning: redefinition of typedef 'FsVerityContext' is a C11 feature [-Wtypedef-redefinition]
   20 | typedef struct FsVerityContext FsVerityContext;
      |                                ^
../libcomposefs/lcfs-internal.h:33:10: note: '../libcomposefs/lcfs-fsverity.h' included multiple times, additional include site here
   33 | #include "lcfs-fsverity.h"
      |          ^
../libcomposefs/lcfs-writer-erofs.c:24:10: note: '../libcomposefs/lcfs-fsverity.h' included multiple times, additional include site here
   24 | #include "lcfs-fsverity.h"
      |          ^
../libcomposefs/lcfs-fsverity.h:20:32: note: unguarded header; consider using #ifdef guards or #pragma once
   20 | typedef struct FsVerityContext FsVerityContext;
      |                                ^
../libcomposefs/lcfs-writer-erofs.c:1618:2: error: call to undeclared function 'static_assert'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 1618 |         static_assert((sizeof(struct erofs_xattr_ibody_header) == LCFS_XATTR_HEADER_SIZE),
      |         ^
1 warning and 1 error generated.

```

it seems that some C11 features are used, but `c_std=c99` is specified in `meson.build`[1], which should be fixed.

[1]: https://github.com/containers/composefs/blob/bebc8e4dcd1ef511d37eabbdcb83f519167fe775/meson.build#L6

Issue #344, opened 2024-09-12
request to relicense CLI tooling from GPLv3 -> dual `GPL-2.0-or-later OR Apache-2.0`


cgwalters:
This is a request to relicense *all* code in this repository under a dual license: `GPL-2.0-or-later OR Apache-2.0` (edit: Clarified from ~GPLv2+ OR ASL2.0~ using the [SPDX terms](https://spdx.org/licenses/)).

Current analysis shows the following users have contributed; I will update this tracker:

- [ ] KatrinJo
- [x] alexlarsson
- [x] allisonkarlitskaya
- [x] bfallik
- [x] cgwalters
- [x] containers (spurious, this is merged branches from the main repo, all from giuseppe)
- [x] eriksjolund
- [x] fboudra
- [x] ffontaine
- [x] giuseppe
- [x] jluebbe
- [x] nekopsykose
- [x] rborn-tx
- [x] wahtari (note: renamed to @divineaugustine ? See https://github.com/containers/composefs/commit/ecef20c which lists as `wahtari` but the PR now shows @divineaugustine )
- [x] wjt

Gathered via this script:

```
git log --merges --pretty=format:'%h %s (%ae)' --full-history -- libcomposefs tools  | sed -e 's,.*from \([-A-Za-z0-9]*\)/.*,\1,' | /bin/sort -u
```

Users with not-small contributions: @eriksjolund @wahtari @jluebbe @rborn-tx @KatrinJo
Smaller sized contributions (1-3 lines) that touch C code: @fboudra @ffontaine @nekopsykose @allisonkarlitskaya 
Smaller sized non-C contributions: @bfallik @wjt

Since I can't ping people necessarily who don't have write access here I'll also go to representative PRs and reuse those as a communication channel.

--- 

Original *obsolete* issue text: 

<details>
Today we have 3 licenses, see https://github.com/containers/composefs/commit/4eaa74f7bae91ee91398503aff209861beca7433

~Right now it's just the CLI tools that are licensed under the GPLv3. The request here is to drop this license from our matrix and relicense the CLI tools as well under LGPLv2+ to just simplify our licensing story.~

At the current time, @alexlarsson and @giuseppe wrote 90% of the initial code there. Any objections?
</details>

giuseppe:
no, I am fine to relicense to LGPLv2+

alexlarsson:
No objections here.

jberkus:
Just to be accurate: you currently have **6** licenses.

* APL2
[...]

PR #365, opened 2024-09-26
meson.build: raise used C standard to C11

ziyao233:
Some C11 features have been used, but meson.build still specifies c_std=c99, leading to errors on Clang compiler.

Fixes: https://github.com/containers/composefs/issues/364

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,7 +3,7 @@ project(
   'c',
   version : '1.0.6',
   default_options : [
-    'c_std=c99',
+    'c_std=c11',
     'warning_level=1',
   ],
 )


cgwalters:
Thanks for this! Actually testing with a somewhat recent clang (`clang version 18.1.8`) here I get compilation failures.

I went ahead and pushed a few additional commits that add a CI check with clang and also fix those errors.

cgwalters:
Also same question as the other PR re licensing: https://github.com/containers/composefs/issues/344
Are you OK with those license terms?

cgwalters (approved):

cgwalters:
I pushed some other changes here so I probably shouldn't merge without a 3rd reviewer.

cgwalters:
Especially the ioctl changes need careful review, I only sanity checked this. Hmm I may just mark as draft until we've done that.

@ziyao233 what version of clang are you using? It must be an older version that doesn't warn about some gnu-specific stuff?

Ohh or maybe the conceptual problem is those are just warnings but in CI we use `-Werror`, and so my attempt to add CI coverage clashed with that. Let me see...perhaps should just take your change as is and I'll split my patches to a separate PR.

ziyao233:
> Especially the ioctl changes need careful review, I only sanity checked this. Hmm I may just mark as draft until we've done that.
> 
> @ziyao233 what version of clang are you using? It must be an older version that doesn't warn about some gnu-specific stuff?

Exactly 18.1.8 and it does throw some warnings, but I didnt come up with a graceful solution to some of them, especially fsverity related things. I decided to split warning fixes into another PR (they are just causing warnings but not error).

> Ohh or maybe the conceptual problem is those are just warnings

Yep

> but in CI we use `-Werror`, and so my attempt to add CI coverage clashed with that. Let me see...perhaps should just take your change as is and I'll split my patches to a separate PR.

seems a good idea to me

PR #366, opened 2024-09-26
libcomposefs: include linux/limits.h to satisfy usage of XATTR_NAME_MAX

ziyao233:
XATTR_NAME_MAX is defined in linux/limits.h, but leaked by glibc since its sys/param.h includes the former header, which is not true on musl and leads to errors like:

```
	../libcomposefs/lcfs-writer.c:1688:16: error: use of undeclared identifier 'XATTR_NAME_MAX'
	 1688 |         if (namelen > XATTR_NAME_MAX) {
```

Include linux/limits.h explicitly to build properly on musl.

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -30,6 +30,7 @@
 #include <errno.h>
 #include <string.h>
 #include <linux/fsverity.h>
+#include <linux/limits.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>


cgwalters (approved):
Hi, thanks for the patch!

First a procedural note, we have an ongoing attempt at a relicensing: https://github.com/containers/composefs/issues/344

Are you OK with licensing this contribution under those terms instead of the current ones?

Also, it should be trivial to add a CI build on a musl system, would be a nice followup to this.

ziyao233:
> Hi, thanks for the patch!
> 
> First a procedural note, we have an ongoing attempt at a relicensing: https://github.com/containers/composefs/issues/344
> 
> Are you OK with licensing this contribution under those terms instead of the current ones?
> 
> Also, it should be trivial to add a CI build on a musl system, would be a nice followup to this.

I'm not familiar with GitHub CI, but if possible I suggest building on Alpine Linux, which is also the most popular musl-based distro.

PR #367, opened 2024-09-27
Add a libcomposefs-internal

cgwalters:
I came across the fact that we had two copy-paste implementations
of converting a binary digest to hex.
    
Deduplicate with an internal shared library.

NOTE: I don't speak meson fluently so please check if I messed
something up there.

### DIFF
--- a/libcomposefs/meson.build
+++ b/libcomposefs/meson.build
@@ -1,3 +1,15 @@
+internal_source_files = files([
+  'lcfs-utils.h',
+  'lcfs-utils.c',
+])
+
+libcomposefs_internal = static_library('composefs-internal',
+  internal_source_files,
+  c_args : composefs_hash_cflags + hidden_visibility_cflags,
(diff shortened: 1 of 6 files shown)


libcomposefs/meson.build
```diff
@@ -1,3 +1,16 @@
+internal_source_files = files([
+  'lcfs-utils.h',
+  'lcfs-utils.c',
+])
+
+libcomposefs_internal = static_library('composefs-internal',
+  internal_source_files,
+  c_args : composefs_hash_cflags + hidden_visibility_cflags,
+  dependencies : libcrypto_dep,
```
alexlarsson, line 9: Does it really need the crypto libs?

cgwalters: Not right now, it easily might in the future...but anyways for now I dropped it.

alexlarsson (approved):

Issue #373, opened 2024-09-29
Testing the file `should-fail-self-hardlink.dump` fails with `mkcomposefs: Invalid embedded NUL character at position 22` 


eriksjolund:
I tried e2afe90d4e58b495b7a5b4a7d3e5d2c92a2d61c8

```
# mkcomposefs --from-file composefs/tests/assets/should-fail-self-hardlink.dump resultfile
mkcomposefs: Invalid embedded NUL character at position 22
```

I had expected to see some error message starting with  `mkcomposefs: Self-referential hardlink`
because of the filename _should-fail-self-hardlink.dump_

cgwalters:
Thanks for trying this out! Fix in https://github.com/containers/composefs/pull/374

However, what would clearly be better here is to enhance the test suite to have a notion of "expected error"...which would be nicer to do if we supported comments in dumpfiles (maybe with an opt-in for new syntax) and then our test suite could embed the expected error like:
```
# expect: Self-referential hardlink
/ 4096 40555 2 0 0 0 0.0 - - - 
/037blockdev 0 @63340777 2 0 0 0 0.0 /037blockdev - -
```

or so.

PR #374, opened 2024-09-29
tests: Fix should-fail-self-hardlink to conform to latest syntax

cgwalters:
This was generated by honggfuzz, and only later did I do a patch which denied embedded NUL characters. Update and clean this up so it fails with the expected error.

Closes: https://github.com/containers/composefs/issues/373


jlebon (approved):
I can confirm this fixes #373.

alexlarsson (approved):

PR #375, opened 2024-10-01
internal: Rewrite byteswapping macros

cgwalters:
The real motivation here is to rewrite the code in order to complete the relicensing, because the person who last touched this didn't respond.

And in general, using `_Generic` here cleans things up anyways.


cgwalters:
Hmm, CI failure here looks maybe like old gcc getting confused by `_Generic`? Let me see...

cgwalters:
I rewrote things to just drop the assertions, they aren't that valuable.

alexlarsson (approved):

PR #376, opened 2024-10-02
Relicense most code under `GPL-2.0-or-later OR Apache-2.0`

cgwalters:
We had a big mess of licenses before; this reduces the set considerably. The updated `COPYING` file summarizes things.

Closes: https://github.com/containers/composefs/issues/344


giuseppe (approved):
LGTM

PR #377, opened 2024-10-02
tests: Add a flow for integration test with fsverity enabled

cgwalters:
integration: Clearly show in logs if we tested with fsverity

---

tests: Add a flow for integration test with fsverity enabled

This was an obvious CI coverage gap.

---


cgwalters:
I'm not a fan of this mess of shell script and python and am quite tempted to RIIR both...and in general because we definitely want to test with both fsverity on and off and some of what we do requires privileges I think in the medium term we should go towards "installed/integration tests" and not unit tests (executed by meson).

But for now, this just ensures we have coverage of the integration test flow with fsverity known to be on.

alexlarsson (approved):

PR #384, opened 2024-10-25
lib: Add expansion comments to header

cgwalters:
Tracing through these xattrs, I found it helpful to have comments with the expansion.


alexlarsson (approved):
lgtm

PR #394, opened 2024-11-06
lib: Add lcfs_fd_measure_fsverity

cgwalters:
lib: Add lcfs_fd_measure_fsverity

Our history with fsverity APIs is a bit messy. For now historical
reasons lcfs_fd_get_fsverity tries to query the kernel (via ioctl)
but will silently fall back to userspace computation - which
is sometimes desirable, other times not.

We also have lcfs_fd_compute_fsverity which is unconditionally
userspace.

However some cases actually really want to require the
fd to have fsverity - so add an API to do that.

---

writer: Canonicalize no-verity errno to -ENOVERITY

This is what we do elsewhere.

---

lib/mount: Use lcfs_fd_measure_fsverity

This is ensuring we have our fsverity ioctl parsing code in
one place.

---

rust: Bind lcfs_fd_measure_fsverity

This is a reasonable thing to want to do.

---

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -576,6 +576,10 @@ int lcfs_fd_measure_fsverity(uint8_t *digest, int fd)
 	fsv->digest_size = MAX_DIGEST_SIZE;
 	int res = ioctl(fd, FS_IOC_MEASURE_VERITY, fsv);
 	if (res == -1) {
+		if (errno == ENODATA || errno == EOPNOTSUPP || errno == ENOTTY) {
+			// Canonicalize errno
+			errno = ENOVERITY;
+		}
 		return -errno;
 	}
 	// The file has fsverity enabled, but with an unexpected different algorithm (e.g. sha512).
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -568,13 +568,15 @@ int lcfs_compute_fsverity_from_fd(uint8_t *digest, int fd)
 // is an error if fsverity is not enabled.
 int lcfs_fd_measure_fsverity(uint8_t *digest, int fd)
 {
-	char buf[sizeof(struct fsverity_digest) + MAX_DIGEST_SIZE];
-	struct fsverity_digest *fsv = (struct fsverity_digest *)&buf;
+	union {
+		struct fsverity_digest fsv;
+		char buf[sizeof(struct fsverity_digest) + MAX_DIGEST_SIZE];
+	} result;
 
 	// First, ask the kernel if the file already has fsverity; if so we just return
 	// that.
-	fsv->digest_size = MAX_DIGEST_SIZE;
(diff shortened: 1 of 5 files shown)


libcomposefs/lcfs-writer.c
```diff
@@ -564,10 +564,9 @@ int lcfs_compute_fsverity_from_fd(uint8_t *digest, int fd)
 	return lcfs_compute_fsverity_from_content(digest, &_fd, fsverity_read_cb);
 }
 
-// Given a file descriptor, first query the kernel for its fsverity digest.  If
-// it is not available in the kernel, perform an in-memory computation.  The file
-// position will always be reset to zero if needed.
-int lcfs_fd_get_fsverity(uint8_t *digest, int fd)
+// Given a file descriptor, query the kernel for its fsverity digest. It
+// is an error if fsverity is not enabled.
+int lcfs_fd_require_fsverity(uint8_t *digest, int fd)
```
allisonkarlitskaya, line 569: I think the better verb here is `measure` to match the name of the ioctl.

Then we'd have:
 - `get`: get it for me, don't care how
 - `compute`: actually do the work in full, in userspace
 - `measure`: use the measure ioctl

cgwalters: Good suggestion, fixed!

cgwalters:
> test-lcfs: ../tests/test-lcfs.c:91: test_no_verity: Assertion `errsv == ENOVERITY' failed.

But only on s390x?? Hmm...

cgwalters:
Hummm I would bet money this is something like the run-on-arch action using qemu's userspace emulation which probably doesn't implement the ioctl so we're getting ENOSYS

cgwalters:
OK, this passes CI now.

libcomposefs/lcfs-writer.c
```diff
@@ -564,10 +564,9 @@ int lcfs_compute_fsverity_from_fd(uint8_t *digest, int fd)
...
-// Given a file descriptor, first query the kernel for its fsverity digest.  If
-// it is not available in the kernel, perform an in-memory computation.  The file
-// position will always be reset to zero if needed.
-int lcfs_fd_get_fsverity(uint8_t *digest, int fd)
+// Given a file descriptor, query the kernel for its fsverity digest. It
+// is an error if fsverity is not enabled.
+int lcfs_fd_measure_fsverity(uint8_t *digest, int fd)
 {
 	char buf[sizeof(struct fsverity_digest) + MAX_DIGEST_SIZE];
 	struct fsverity_digest *fsv = (struct fsverity_digest *)&buf;
```
smcv, line 572: Drive-by comment: I think this has the same alignment issue that's fixed by https://github.com/ostreedev/ostree/pull/3340

cgwalters: Yep, while this PR is orthogonal conceptually (just moving the code) I've rolled in the same fix here because otherwise it'd conflict. Thanks!

jluebbe (approved):

PR #395, opened 2024-11-06
rust: Bind our default recommended max inline size

cgwalters:
It's quite important that this gets used consistently for interoperability - specifically, what would be really good is to ensure that going from "tar -> composefs" has a single well known digest. The choice of file inlining means different composefs projects can disagree on this.

Let's encourage this to be part of the composefs standard.


cgwalters:
Closing in favor of https://github.com/containers/composefs-rs/issues/26

PR #400, opened 2024-12-02
writer: Reject empty xattr names

cgwalters:
Came out of a related discussion in ostree
https://github.com/ostreedev/ostree/pull/3346#discussion_r1866007266

Today we faithfully write this, and `fsck.erofs` doesn't seem to complain...but such a thing makes no sense so we shouldn't support it.

Actually digging in more here...in practice the way xattrs work in Linux at least is they're restricted to starting with one of the standard prefixes like `security.` or `user.`; but enforcing/validating that is a whole other thing. For now let's just deny this edge case.


allisonkarlitskaya (approved):
Looks good.

Do we try to cover the duplicated case?

cgwalters:
> Do we try to cover the duplicated case?

The semantics for that today are that you get last-one-wins. We hit a bug related to this when fuzzing, see https://github.com/containers/composefs/commit/2ef9e9bd9d65b95d3f7c9d22da5712a4bfe46a6a

We could probably very safely change that to be an error? It'd make sense to me to do so.

PR #401, opened 2024-12-02
writer: Fix return value of lcfs_node_unset_xattr

cgwalters:
Since the first creation of this code in
5ac1f5c4 "lib: Update xattr APIs" this function
has always returned an error code - but nothing
checked it, so it didn't matter.

Now also, currently this function cannot fail
but let's give ourselves some flexibility here;
perhaps we want to e.g. invoke `realloc`
and in that case we'd need to handle OOM.

I just noticed this while working on commit
02077e89b5b81c76e962a6643c7a7423b1f8336f
to reject empty xattr names.

Change this to return success, and also check its
return value in the set path.

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -1660,24 +1660,26 @@ int lcfs_node_unset_xattr(struct lcfs_node_s *node, const char *name)
 {
 	ssize_t index = find_xattr(node, name);
 
-	if (index >= 0) {
-		struct lcfs_xattr_s *xattr = &node->xattrs[index];
-		size_t value_len = xattr->value_len;
-		free(xattr->key);
-		free(xattr->value);
-		if (index != (ssize_t)node->n_xattrs - 1)
-			node->xattrs[index] = node->xattrs[node->n_xattrs - 1];
-		node->n_xattrs--;
-		// Note 2*size - to account for worst case alignment
-		if (node->n_xattrs > 0)
-			node->xattr_size -= (2 * LCFS_INODE_XATTRMETA_SIZE - 1) +
-					    strlen(name) + value_len;
-		else
-			node->xattr_size = 0; // If last xattr, remove the overhead too
-		assert(node->xattr_size >= 0);
+	if (index < 0) {
+		errno = ENODATA;
+		return -1;
 	}
 
-	return -1;
+	struct lcfs_xattr_s *xattr = &node->xattrs[index];
+	size_t value_len = xattr->value_len;
+	free(xattr->key);
+	free(xattr->value);
+	if (index != (ssize_t)node->n_xattrs - 1)
+		node->xattrs[index] = node->xattrs[node->n_xattrs - 1];
+	node->n_xattrs--;
+	// Note 2*size - to account for worst case alignment
+	if (node->n_xattrs > 0)
+		node->xattr_size -= (2 * LCFS_INODE_XATTRMETA_SIZE - 1) +
+				    strlen(name) + value_len;
+	else
+		node->xattr_size = 0; // If last xattr, remove the overhead too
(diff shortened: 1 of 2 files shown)


libcomposefs/lcfs-writer.c
```diff
@@ -1677,7 +1677,7 @@ int lcfs_node_unset_xattr(struct lcfs_node_s *node, const char *name)
 		assert(node->xattr_size >= 0);
 	}
 
-	return -1;
+	return 0;
```
allisonkarlitskaya, line 1680: I think this change is wrong.  This -1 is the case where we didn't find the xattr to remove.  Arguably we should set `errno = ENODATA` here (which is the same as the `removexattr()` syscall.

The thing that's missing is the `return 0` on the "good path" inside the `if() { }` block, just above.

allisonkarlitskaya (changes requested):
Something went wrong with the commit message formatting...

libcomposefs/lcfs-writer.c
cgwalters: > I think this change is wrong. 

I think it's the original code that's wrong.

>  This -1 is the case where we didn't find the xattr to remove. Arguably we should set errno = ENODATA here (which is the same as the removexattr() syscall.
> The thing that's missing is the return 0 on the "good path" inside the if() { } block, just above.

The API you propose makes sense to me and it does seem theoretically valuable for a caller to be able to distinguish "value not found" from other errors. But I'd note that it complicates the desired logic for the only current caller of this from our own code.

But I'd also say my proposed change is not "wrong"; it has pretty defensible API semantics in that just changes the function to work the same as `lcfs_node_set_xattr` does basically (which also currently does not signal to the caller in any way whether there was an extant value).

Anyways though since I don't have a strong feeling on this I'll change it to match your suggestion, thanks!

allisonkarlitskaya: > I think it's the original code that's wrong.

To be clear: I agree :)

cgwalters: OK, I changed it.

Though one thing this change makes really obvious is that we desperately need API documentation using a standard tool that does things like describe semantics and especially return error codes. I'm out of the loop on state of the art in tooling there...systemd seems to do man-page-per-function which seems quite heavyweight but maybe is viable for us. Is doxygen still the top thing? (Seems so?)

cgwalters:
I also added a test case for the multiple xattr path.

cgwalters:
> Something went wrong with the commit message formatting...

Can you be more specific?

allisonkarlitskaya:
> Can you be more specific?

Same as we talked about today.  It's wrapped at 50.  Not important :)

libcomposefs/lcfs-writer.c
allisonkarlitskaya: What ever happened to good old gtk-doc? :)

libcomposefs/lcfs-writer.c
```diff
@@ -1660,24 +1660,26 @@ int lcfs_node_unset_xattr(struct lcfs_node_s *node, const char *name)
...
-			node->xattrs[index] = node->xattrs[node->n_xattrs - 1];
-		node->n_xattrs--;
-		// Note 2*size - to account for worst case alignment
-		if (node->n_xattrs > 0)
-			node->xattr_size -= (2 * LCFS_INODE_XATTRMETA_SIZE - 1) +
-					    strlen(name) + value_len;
-		else
-			node->xattr_size = 0; // If last xattr, remove the overhead too
-		assert(node->xattr_size >= 0);
+	if (index < 0) {
```
allisonkarlitskaya, line 1663: This way around reads a lot nicer now.  Thanks for the change.

allisonkarlitskaya (approved):
Thanks!

PR #402, opened 2024-12-14
Add release binaries from git main on push

cgwalters:
Targeting a few distributions, produce a tarball with binaries.

This is intended to be used in other downstream CI scenarios.

### DIFF
--- a/.github/workflows/builds.yaml
+++ b/.github/workflows/builds.yaml
@@ -1,0 +1,55 @@
+name: Build binaries
+
+on: push
+
+permissions:
+  actions: read
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        include:
+          - baseimage: docker.io/library/ubuntu:24.04
+            basename: ubuntu-24.04
+          - baseimage: docker.io/library/ubuntu:22.04
+            basename: ubuntu-22.04
+          - baseimage: quay.io/fedora/fedora:41
+            basename: fedora-41
+          - baseimage: quay.io/centos/centos:stream9
+            basename: centos-stream9
+    runs-on: ubuntu-24.04
+    container: ${{ matrix.baseimage }}
+    steps:
+      - name: Bootstrap git
+        run: if test -x /usr/bin/apt; then apt -y update && apt -y install git; else dnf -y install git; fi
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Fix git perms
+        run: git config --global --add safe.directory $(pwd)
+      - name: Install dependencies
+        run: ./hacking/installdeps.sh
+      - name: Set environment
+        run: set -e
+             echo "SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)" >> $GITHUB_ENV
+      - name: Configure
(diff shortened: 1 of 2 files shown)


.github/workflows/builds.yaml
```diff
@@ -0,0 +1,55 @@
+name: Build binaries
+
+on: push
+
+permissions:
```
allisonkarlitskaya, line 5: I didn't know that this could appear at the toplevel!  Interesting.

allisonkarlitskaya: I've usually used `permissions: {}` for this in the past, but I've also been setting it per-job, so I don't know if it's the same...

hacking/installdeps.sh
```diff
@@ -33,9 +48,11 @@ done
 
 # Install packages:
 if [ -n "${PACKAGES_REQUIRED}" ]; then
+    apt -y update
```
allisonkarlitskaya, line 51: Probably makes sense to do this once, regardless.  We know that one set or the other will be non-empty (if not both).

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Capture build
+        run: set -e; DESTDIR=$(pwd)/instroot meson install -C build &&
+             tar -C instroot --sort=name --owner=0 --group=0 --numeric-owner
+               --mtime="${SOURCE_DATE_EPOCH}"
+               --pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime
+               -czf composefs.tar.gz .
+      - name: Upload binary
+        uses: actions/upload-artifact@v4
+        with:
+          name: composefs-${{ matrix.basename }}.tar
```
allisonkarlitskaya, line 48: Is it too difficult to derive this from `/usr/lib/os-release` of the running container?

.github/workflows/builds.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Build
+        run: meson compile -C build
+      - name: Capture build
+        run: set -e; DESTDIR=$(pwd)/instroot meson install -C build &&
+             tar -C instroot --sort=name --owner=0 --group=0 --numeric-owner
+               --mtime="${SOURCE_DATE_EPOCH}"
+               --pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime
+               -czf composefs.tar.gz .
+      - name: Upload binary
+        uses: actions/upload-artifact@v4
```
allisonkarlitskaya, line 46: From my memory, artifacts have a fairly limited usefulness in terms of lifetime and not being accessible unless you have an API token or are logged in.

What usecase are you imagining for these?  Plucked from a workflow running in a different repository?  How would it find them?

allisonkarlitskaya (approved):
I think this is fine to go in.  In my opinion the greatest benefit here is that we get exposure to building on different distributions.  Maybe you'd like to run the tests as well?

I'm not sure about how the artifacts themselves might end up being used.  It might be more interesting if we did those as part of our release process and attached them to the release where it would be easier to download them.

Issue #361, opened 2024-09-26
mount: add support for optionally skipping loopback device


allisonkarlitskaya:
This is cool: https://lore.kernel.org/lkml/CAMuHMdVqa2Mjqtqv0q=uuhBY1EfTaa+X6WkG7E2tEnKXJbTkNg@mail.gmail.com/T/

allisonkarlitskaya:
The complexity of the mount code is growing a lot:
 - maybe we opened the fd by filename or got it from the user
 - maybe we get an extra fd because of the memfd copy
 - maybe we get an extra fd because of loopback devices (or in the future, maybe not)

So this storing the `fd` in the `state` thing is probably not gonna survive...  There's just too many fds now...

The thing about opening vs. getting an fd from the user is pretty boring and is well handled in the two separate entrypoint APIs for each of those cases, but then we need to go on a path:
 - replace that fd with a memfd, if appropriate
 - replace that fd with a loop device if appropriate
 - mount that

plus all the cleanups of optionally closing fds if we were the ones that opened it.  I start to think that maybe we could dup() the user's fd just to simply things there (ie: we always close it) and close() each fd from a helper before we return the "next layer" one.

Basically, the handling of trying to make sure we rmdir() and close() stuff as appropriate is getting a bit too complicated... particularly when mixed with various early error exits...

alexlarsson:
This is definitely cool, and will help tons to avoid loads of loopback devices e.g. in the container image case. I agree that the mount code complexity is growing, in particular since we need to support legacy mount APIs, etc. But, this need only be included in the modern API mount codepath, and I believe the complexity is worth it.

allisonkarlitskaya:
@alexlarsson what's the story behind needing to support the legacy mount APIs?

alexlarsson:
@allisonkarlitskaya Its needed on e.g. rhel9 at least I belive.

cgwalters:
As I understand things today, if we go with the current design in https://github.com/containers/composefs/issues/332 then we don't need the memfd path, right?

Skipping the loopback though would be quite valuable down the line, but is IMO not critical right now again versus just fleshing out the status quo.

>  I start to think that maybe we could dup() the user's fd just to simply things there (ie: we always close it) and close() each fd from a helper before we return the "next layer" one.

That seems fine to me if it simplifies the code though.

allisonkarlitskaya:
[...]

Issue #144, opened 2023-06-08
better loopback handling (hiding it)


cgwalters:
After a lot of debate, it seems like we will be focusing on the "erofs+overlayfs" flow.  There are positives and negatives to this.

This issue is about one of the negative things we lose with this combination, which is that we need to make a loopback device.

In our usage, the loopback device is an implementation detail of "composefs".  However, its existence leaks out to all of the rest of the system, e.g. it shows up in `lsblk`, there's objects in `/sysfs` for it, etc.

One thing I'd bikeshed here is that perhaps using the [new mount API](https://lwn.net/Articles/753473/) we could add something like this

```
diff --git a/libcomposefs/lcfs-mount.c b/libcomposefs/lcfs-mount.c
index ea2c2e9..b9d608d 100644
--- a/libcomposefs/lcfs-mount.c
+++ b/libcomposefs/lcfs-mount.c
@@ -393,7 +393,7 @@ static int lcfs_mount_erofs(const char *source, const char *target,
                return -errno;
        }
 
-       res = syscall_fsconfig(fd_fs, FSCONFIG_SET_STRING, "source", source, 0);
+       res = syscall_fsconfig(fd_fs, FSCONFIG_SET_FD, "loop-file", src_fd, 0);
        if (res < 0)
                return -errno;
```
So instead of passing the `/dev/loopX` pathname, we just give an open fd to the kernel (to erofs) and internally it creates the loopback setup.  But the key here is that this block device would be exclusively owned by the erofs instance, it wouldn't be visible to userspace.

hsiangkao:
Also, directly reading files rather than block devices could be done in principle for erofs as the previous kcomposefs (and it's useful for all fses since all fses have use cases to use loopbacked mounts, and I could even duplicate a whole fscache caching framework so that it's more flexiable), but the problem is that rare in-kernel filesystems really work in this way.

It also duplicates iomap interface. If we'd like to avoid loopback devices, I'd suggest if we could support file-based backend in addition to block devices in iomap and make fscache work with iomap so I could cleanup the current erofs codebase as well.

alexlarsson:
Yeah, having generic direct-file via the VFS for all filesystems that support iomap would be great.

cgwalters:
So to me, the new "composefs" is about putting together things that already exist (overlayfs, fs-verity, erofs for metadata etc.).

There's actually precedent for efficient in-kernel-only access to a file: and that's *swap files*.  The more I think about it, the more strong this alignment is:

[...]

PR #403, opened 2024-12-19
mount: Support EROFS file-backed mounts

ruihe774:
It turns out we can directly mount `/dev/fd/$fd`, so current API does not need to be changed.

Closes: #361
Closes: https://github.com/containers/composefs/issues/144

### DIFF
--- a/libcomposefs/lcfs-mount.c
+++ b/libcomposefs/lcfs-mount.c
@@ -243,7 +243,7 @@ static errint_t setup_loopback(int fd, const char *image_path, char *loopname)
 		return -errsv;
 	}
 
-	sprintf(loopname, "/dev/loop%ld", devnr);
+	snprintf(loopname, PATH_MAX, "/dev/loop%ld", devnr);
 	loopfd = open(loopname, O_RDWR | O_CLOEXEC);
 	if (loopfd < 0)
 		return -errno;
@@ -578,16 +578,12 @@ static errint_t lcfs_mount_erofs_ovl(struct lcfs_mount_state_s *state,
 	char imagemountbuf[] = "/tmp/.composefs.XXXXXX";
 	char *imagemount;
 	bool created_tmpdir = false;
-	char loopname[PATH_MAX];
+	char source[PATH_MAX];
 	int errsv;
 	errint_t err;
 	int loopfd;
 
 	image_flags = lcfs_u32_from_file(header->flags);
-
-	loopfd = setup_loopback(state->fd, state->image_path, loopname);
-	if (loopfd < 0)
-		return loopfd;
 
 	if (options->image_mountdir) {
 		imagemount = (char *)options->image_mountdir;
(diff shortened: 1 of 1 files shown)


libcomposefs/lcfs-mount.c
```diff
@@ -578,34 +578,40 @@ static errint_t lcfs_mount_erofs_ovl(struct lcfs_mount_state_s *state,
...
 			errsv = errno;
-			close(loopfd);
 			return -errsv;
 		}
 		created_tmpdir = true;
 	}
 
-	err = lcfs_mount_erofs(loopname, imagemount, image_flags, state);
-	close(loopfd);
+	sprintf(source, "/dev/fd/%d", state->fd);
```
allisonkarlitskaya, line 599: `snprintf` please, even if it's "impossible" to overflow.

We usually say `/proc/self/fd/%d` there.

I'd be curious about how this shows up in `/proc/mounts`.  It's a purely theoretical question, though, since we unmount it very quickly.

libcomposefs/lcfs-mount.c
```diff
@@ -578,34 +578,40 @@ static errint_t lcfs_mount_erofs_ovl(struct lcfs_mount_state_s *state,
...
+	err = lcfs_mount_erofs(source, imagemount, image_flags, state);
+
 	if (err < 0) {
```
allisonkarlitskaya, line 602: Please check this for the specific `errno` we're expecting if we do this on older kernels or ones with the "mount file-backed" feature disabled.  Probably something like `ENOTBLK` or so?

allisonkarlitskaya (changes requested):
Using this feature would be quite nice.  `composefs-rs` does it only this way (with no support for older kernels).

Here's a couple of immediate issues I noticed:

libcomposefs/lcfs-mount.c
ruihe774: I don't think `sprintf` is a problem. `setup_loopback` uses it as well:
https://github.com/containers/composefs/blob/1003b4f20c93a3475ca4fc028c3396870a58db63/libcomposefs/lcfs-mount.c#L246

It's impossible the length of `/dev/fd/` plus the length of `%d` exceeds `PATH_MAX`.

allisonkarlitskaya: I agree it's not a real problem, and the readlink thing would be pretty gross.

allisonkarlitskaya: The `sprintf` thing is a problem, though, and we should probably fix the other user as well.

ruihe774: > The `sprintf` thing is a problem, though, and we should probably fix the other user as well.

Okay then

cgwalters (approved):
Thanks!

cgwalters:
I'm marking this as closing both of the outstanding issues we have around loopback.

cgwalters (approved):
Thanks!

cgwalters:
Hm, technically the merge settings right now require the person who marked things as "requested changes" to switch to approval but I can see you've addressed those changes.

PR #407, opened 2025-01-06
The spec file refers to missing license files

jnovy:
The spec file refers to missing license files making making the upstream spec unusable. Also adding missing License tag.

### DIFF
--- a/composefs.spec.in
+++ b/composefs.spec.in
@@ -36,6 +36,7 @@ Devel files for %{name}.
 
 %package        libs
 Summary:        Libraries for %{name}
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
I'd still like to get rid of this upstream spec file as in my experience it never works trying to keep it in sync across multiple distros, as is happening here.

jnovy:
Thanks for merging this @cgwalters I still think there is a value in maintaining the spec file for RPM based distros. We do the same for podman, etc. I volunteer to maintain this for you if you are cool with it.

PR #410, opened 2025-01-24
Some changes for format 1.2

allisonkarlitskaya:
This is very incomplete but it's enough to get mkcomposefs to output simple images that are identical to the images coming from the (also modified) erofs writer code in composefs-rs.

I tried to keep individual changes as individual commits so we can :+1:/:-1: each change independently.

We're going to want to add some more stuff around handling of alignment of symlinks and such (which no longer needs to be treated as a special case as of kernel 6.12) and probably some other points.

See #198

### DIFF
--- a/libcomposefs/lcfs-writer-erofs.c
+++ b/libcomposefs/lcfs-writer-erofs.c
@@ -992,9 +992,10 @@ static int write_erofs_inode_data(struct lcfs_ctx_s *ctx, struct lcfs_node_s *no
 		for (size_t i = 0; i < node->n_xattrs; i++) {
 			struct lcfs_xattr_s *xattr = &node->xattrs[i];
 			if (xattr->erofs_shared_xattr_offset >= 0) {
-				uint64_t offset =
-					ctx_erofs->inodes_end % EROFS_BLKSIZ +
-					xattr->erofs_shared_xattr_offset;
+				uint64_t offset = ctx_erofs->inodes_end;
+				if (ctx->options->version < 2)
+					offset %= EROFS_BLKSIZ;
+				offset += xattr->erofs_shared_xattr_offset;
 				uint32_t v = lcfs_u32_to_file(
 					(uint32_t)(offset / sizeof(uint32_t)));
 				ret = lcfs_write(ctx, &v, sizeof(v));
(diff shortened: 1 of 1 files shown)


libcomposefs/lcfs-writer.h
```diff
@@ -37,10 +37,11 @@ enum lcfs_flags_t {
 	LCFS_FLAGS_MASK = 0,
 };
 
-#define LCFS_VERSION_MAX 1
+#define LCFS_VERSION_MAX 2
```
cgwalters, line 40: Mechanically I think it'd be helpful for us to add the concept of an "experimental" version. And we probably only want at most one of those at a time.

So how about
`#define LCFS_VERSION_EXPERIMENTAL 2`

And on the CLI it's `mkcomposefs --format-version=experimental` and in the C API maybe we can at least require them to set something like `acknowledge_experimental: bool` in `struct lcfs_write_options_s`?

Basically if we just clearly flag this as experimental I don't see any problems in merging to main and iterating from there, right?

cgwalters:
Also on mechanics, I think we'll need to ensure the integration tests cover all format versions, and the fuzzer path. The fuzzer one in particular adds a lot of coverage nowadays as it ensures we can also read back what we wrote.

libcomposefs/lcfs-writer.h
allisonkarlitskaya: That makes sense.  I think I'd propose a separate `--experimental` flag in that case, though, since it would likely be easier to implement (since the current version values are expected to be integers).

allisonkarlitskaya: I just thought about a potential annoying issue here (although I think it's an issue we'll have anyway): the composefs-rs testsuite calls `mkcomposefs` (soon with `--experimental`) to compare that the outputs are the same.  If we expect `--experimental` to shift meaning in the future then this could get annoying....

cgwalters: I think we should also add `mkcomposefs --version` (on general principle) and have it output in yaml format (docker, `podman version`, ostree do this):

```
mkcomposefs:
  version: 1.0.7
  experimental-format: 2.0 # like a sub-version in the experimental format
```

Then the test suite can at least cover the installed mkcomposefs's version; we basically skip in the unit/integration tests unsupported versions. In CI (i.e. GHA) for composefs-rs we'd install a pinned version of composefs by default etc.

alexlarsson: The problem we have is that we have a goal in mind, where we have a series of increasing version numbers, that have well defined, non-changing meaning. But we also want that to be able to merge partial work on new features that will then eventually combine into the new fully supported version.

So, --experimental would be an opt-in to getting a potentially partial new version. However, I don't really like that we'd have a switch that produces output that varies over time.

Can we maybe introduce a new version number for the currently in-progress "next" version. Say, lets call the "next stable version" 2, and we give the current, in-tree experimental version for it "1". To build such an image you'd have to put in --format-version=2 --experimental-version=1. Whenever we make an unstable version format change we bump this, and now users have to pass --format-version=2 --experimental-version=2. Until eventually the version is stable. (Note: We would only support the latest experimental version, others would just fail).

libcomposefs/lcfs-writer-erofs.c
```diff
@@ -992,9 +992,10 @@ static int write_erofs_inode_data(struct lcfs_ctx_s *ctx, struct lcfs_node_s *no
 		for (size_t i = 0; i < node->n_xattrs; i++) {
 			struct lcfs_xattr_s *xattr = &node->xattrs[i];
 			if (xattr->erofs_shared_xattr_offset >= 0) {
-				uint64_t offset =
-					ctx_erofs->inodes_end % EROFS_BLKSIZ +
-					xattr->erofs_shared_xattr_offset;
+				uint64_t offset = ctx_erofs->inodes_end;
+				if (ctx->options->version < 2)
+					offset %= EROFS_BLKSIZ;
+				offset += xattr->erofs_shared_xattr_offset;
```
alexlarsson, line 998: I'm not particularly fond of how this conditional is done. It makes it quite complicated to understand what is going on. I think a more plain if (version < 2) { do it in old way} else { do it in new way } would be easier to understand.

alexlarsson:
I gotta say that I'm not overly sold that this small set of changes make any real drastic difference to the file format, but I'm also not against any individual change.

libcomposefs/lcfs-writer.h
alexlarsson: Or, to be less painful, maybe we can call it "--format-version=2beta1", which we then bump to "--format-version=2beta2", and eventually "2", when it is stablilized.

allisonkarlitskaya:
These changes aren't about "big wins in terms of efficiency" or anything like that, but more about "reducing the complexity of the implementation" with an eye towards being able to succinctly document the process of creating a composefs image in a way that a 3rd party implementation would have a fighting chance of being bit-for-bit compatible with the existing writer implementations.

alexlarsson:
Maybe we should look into getting this landed. One thing I dislike about this is that we break the backwards compat, but we don't take advantage of the custom xattr prefix, which I think will help a lot keeping the erofs image size down.

Unfortunately that is not supported in RHEL9, but RHEL9 should be able to use the old format.

PR #412, opened 2025-03-10
man: fix section headers

rohieb:
While packaging for Debian, lintian complained about the section headers:

```
W: composefs: wrong-manual-section 1 != composefs [usr/share/man/man1/composefs-info.1.gz:2]
W: composefs: wrong-manual-section 1 != composefs [usr/share/man/man1/mkcomposefs.1.gz:2]
W: composefs: wrong-manual-section 1 != composefs [usr/share/man/man1/mount.composefs.1.gz:2]
W: composefs: wrong-manual-section 5 != composefs [usr/share/man/man5/composefs-dump.5.gz:2]
```

(See <https://lintian.debian.org/tags/wrong-manual-section.html> for full info about the warning.)

According to man-pages(7), the .TH syntax is:

```
.TH title section date source manual-section
```

Remove the parentheses so that the section is detected correctly, leave the date empty, and escape the space in the last argument to comform to the required syntax.

### DIFF
--- a/man/composefs-dump.md
+++ b/man/composefs-dump.md
@@ -1,4 +1,4 @@
-% composefs-dump(5) composefs | User Commands
+% composefs-dump 5 "" composefs "User Commands"
 
 # NAME
 
--- a/man/composefs-info.md
+++ b/man/composefs-info.md
@@ -1,4 +1,4 @@
-% composefs-info(1) composefs | User Commands
+% composefs-info 1 "" composefs "User Commands"
 
 # NAME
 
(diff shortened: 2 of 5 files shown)


cgwalters (approved):
Only giving this a superficial glance but looks sane to me, thanks!

cgwalters:
Hmm looks like CI fell over; segfault on ppc64le setting up the packages; probably a qemu bug, although it could be a userspace bug too. We should at least be able to use the native aarch64 runners.

For the ASAN one, mmm we're going to need to extract core dumps there.

rohieb:
Lintian found another thing related to the manpages: since mount.composefs is in /usr/sbin, it should live in section 8 ("System administration commands") instead of 1 (userspace tools). I've pushed a respective fix.

PR #415, opened 2025-04-01
tests: ignore EOPNOTSUPP in in fsverity tests

katexochen:
When executing the tests in a sandbox, this is the error we get when running on a kernel with fsverity support, but sandbox restrictions preventing us from mounting something.

I failed to understand why this occurred the first time with v1.0.8.


cgwalters:
Mmmm. I'm going to merge this but I'd prefer really that we instead try to just require execution environments that have fsverity enabled.

cgwalters (approved):

katexochen:
> that we instead try to just require execution environments that have fsverity enabled.

the problem is that fsverity _is_ enabled bit mounting is forbidden

PR #434, opened 2026-03-17
composefs-info: Write 0 for directory sizes in dump output

cgwalters:
composefs-dump(5) already specifies that the SIZE field "is ignored for directories", but composefs-info dump was writing the on-disk EROFS directory data size (e.g. the byte count of serialized erofs_dirent structures). This value is an EROFS implementation detail - POSIX just says st_size is implementation defined.

The motivation here is I want to have tests that round trip from composefs dumpfile -> EROFS -> dumpfile, and currently we discard the input size.

While we continue to ignore the size on dumpfile input, let's change our output to zero, and also suggest that 0 is provided as input.

Assisted-by: OpenCode (Claude Opus 4)

### DIFF
--- a/man/composefs-dump.md
+++ b/man/composefs-dump.md
@@ -53,7 +53,8 @@ otherwise specified):
     line.
 
 **SIZE**
-:   The size of the file. This is ignored for directories.
+:   The size of the file. For directories this should be 0 (it is
+    ignored when reading).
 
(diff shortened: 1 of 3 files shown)


cgwalters:
Motivation: Work on https://github.com/containers/composefs-rs/pull/225

PR #436, opened 2026-05-19
lcfs-writer-erofs: reject hardlinked whiteout inodes with EINVAL

cgwalters:
A whiteout (chardev with rdev=0) represents the absence of a file, so it is semantically invalid for it to have nlink>1.

This was found via fuzzing in composefs-rs, currently we get a heap use-after-free when parsing such files.

Assisted-by: OpenCode (Claude Sonnet 4.6)

### DIFF
--- a/libcomposefs/lcfs-writer-erofs.c
+++ b/libcomposefs/lcfs-writer-erofs.c
@@ -1766,7 +1766,17 @@ static struct lcfs_node_s *lcfs_build_node_from_image(struct lcfs_image_data *da
 	}
 
 	if (type == S_IFCHR && node->inode.st_rdev == 0) {
-		errno = ENOTSUP; /* Use this to signal that we found a whiteout */
+		/* A whiteout (chardev rdev=0) with nlink>1 is semantically
+		 * invalid: a whiteout represents the absence of a file, so it
+		 * cannot be hard-linked.  Reject the image explicitly rather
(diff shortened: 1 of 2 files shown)


libcomposefs/lcfs-writer-erofs.c
```diff
@@ -1766,7 +1766,17 @@ static struct lcfs_node_s *lcfs_build_node_from_image(struct lcfs_image_data *da
 	}
 
 	if (type == S_IFCHR && node->inode.st_rdev == 0) {
-		errno = ENOTSUP; /* Use this to signal that we found a whiteout */
+		/* A whiteout (chardev rdev=0) with nlink>1 is semantically
+		 * invalid: a whiteout represents the absence of a file, so it
+		 * cannot be hard-linked.  Reject the image explicitly rather
+		 * than silently skipping what would be a dangling alias. */
+		if (node->inode.st_nlink > 1) {
```
jeckersb, line 1773: Most minor of nits, should be able to dedupe the `free(...)` calls

```suggestion
        free(hash_remove(data->node_hash, &ht_entry));
        if (node->inode.st_nlink > 1) {
            errno = EINVAL;
            return NULL;
        }
        errno = ENOTSUP;
        return NULL;
```

jeckersb (approved):

PR #445, opened 2026-07-28
lcfs-fsverity: fix root_hash for zero-length files

### LINKED ISSUE #442: composefs-info measure-file reports a wrong fs-verity digest for empty files
`lcfs_fsverity_context_get_digest()` computes a digest for zero-length input that disagrees with the kernel and with fsverity-utils and composefs-rs (composefs-rs's drop-in reimplementation of this very command).

## Reproduction

```console
$ truncate -s 0 empty

$ fsverity digest empty
sha256:3d248ca542a24fc62d1c43b916eae5016878e2533c88238480b26128a1f1af95 empty

$ cfsctl composefs-info measure-file empty   # composefs-rs reimplementation
3d248ca542a24fc62d1c43b916eae5016878e2533c88238480b26128a1f1af95

$ composefs-info measure-file empty          # this repo
cfc3391077edfd51859f6e5e16e03f82b1da3ee1a35d418b0d0b3b57740b4405
```

## Impact

Narrow in practice: composefs images don't reference empty backing objects, so image digests are unaffected. But `composefs-info measure-file` is useful as a conformance test for independent implementations, and there
it reports a digest the kernel can never produce; the same applies to any direct consumer of the `lcfs_fsverity_context_*` API with zero-length input.

cgwalters:
lcfs_fsverity_context_get_digest() unconditionally zero-padded and hashed a full Merkle tree block to produce root_hash, even for zero-length files. Per the fs-verity on-disk format, an empty file has zero Merkle tree blocks, so root_hash must be all-zero bytes instead.

This meant composefs-info measure-file (and any direct caller of lcfs_fsverity_context_get_digest with empty input) reported a digest the kernel, fsverity-utils, and composefs-rs's independent reimplementation can never produce, even though composefs images themselves are unaffected since they never reference empty backing objects.

Assisted-by: https://github.com/cgwalters/cgwalters#llms
Closes: #442

### DIFF
--- a/libcomposefs/lcfs-fsverity.c
+++ b/libcomposefs/lcfs-fsverity.c
@@ -432,8 +432,6 @@ void lcfs_fsverity_context_get_digest(FsVerityContext *ctx,
 {
 	struct fsverity_descriptor descriptor;
 
-	lcfs_fsverity_context_flush_level(ctx, 0);
-
 	memset(&descriptor, 0, sizeof(descriptor));
 	descriptor.version = 1;
 	descriptor.hash_algorithm = 1;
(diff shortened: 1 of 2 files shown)


jeckersb (approved):
👍

PR #446, opened 2026-07-28
mkcomposefs: Add --hardlinks

### LINKED ISSUE #444: mkcomposefs: building from a directory silently loses hardlinks
When building from a directory (`mkcomposefs <dir> <out>`), hardlinked files are emitted as independent inodes. It leads to two consequences:
1. the image's fs-verity digest differs from what any hardlink-aware builder produces for the same tree (image identity depends on the toolchain);

2. the mounted image misreports the mount dir/tree size (and nlink=1, distinct inodes).

This issue is reproducible on v1.0.8 and current main (on 07/28/2026).

The steps to reproduce it:

```
$ mkdir tree && dd if=/dev/zero of=tree/big bs=1M count=100 && ln tree/big tree/big-link
$ du -sh --apparent-size tree
100M	tree

$ mkcomposefs --digest-store=objects --min-version=1 --max-version=1 tree tree.cfs

$ composefs-info dump tree.cfs
/ 4096 40755 2 ...
/big 104857600 100644 1 ... 62/37f0e7... - 6237f0e7...
/big-link 104857600 100644 1 ... 62/37f0e7... - 6237f0e7...

$ fsverity digest tree.cfs
sha256:88d2ef4f83caccd657e6e6844144ffa67dc155a8fe3476e4b8859d5d05643f7c tree.cfs

$ mkdir mnt && mount -t composefs -o basedir=objects tree.cfs mnt

$ du -sh --apparent-size mnt/
200M	mnt/
```

The hardlink is gone: two full entries, nlink=1 each.

composefs-rs's `cfsctl`, reading the *same directory* into the same V1 format, disagrees:
```
$ cfsctl --no-repo --hash sha256 --erofs-version 1 compute-id --no-propagate-usr-to-root tree
18c06b7c0b8cc959d23f2e2a7e1fccb2f402f6cc6052e936be4ee41686127015
```
(The flags only remove unrelated differences: cfsctl otherwise defaults to SHA-512 and propagates "/usr" metadata to the root inode.)

Control: replace the `ln` with `cp` and both tools emit the **identical** digest, the writers agree byte-for-byte at baseline, so the hardlink is the only issue here. This breaks any digest-based cross-check between images built from a directory and by hardlink-aware tools like `composefs-rs`, or `mkcomposefs --from-file`).

cgwalters:
Closes: https://github.com/composefs/composefs/issues/444

### DIFF
--- a/libcomposefs/lcfs-writer.c
+++ b/libcomposefs/lcfs-writer.c
@@ -786,7 +786,7 @@ struct lcfs_node_s *lcfs_load_node_from_file(int dirfd, const char *fname,
 	if (buildflags & ~(LCFS_BUILD_SKIP_XATTRS | LCFS_BUILD_USE_EPOCH |
 			   LCFS_BUILD_SKIP_DEVICES | LCFS_BUILD_COMPUTE_DIGEST |
 			   LCFS_BUILD_NO_INLINE | LCFS_BUILD_USER_XATTRS |
-			   LCFS_BUILD_BY_DIGEST)) {
+			   LCFS_BUILD_BY_DIGEST | LCFS_BUILD_TRACK_HARDLINKS)) {
 		errno = EINVAL;
 		return NULL;
 	}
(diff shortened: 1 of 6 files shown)


libcomposefs/lcfs-writer.c
```diff
@@ -1504,9 +1504,112 @@ bool lcfs_node_dirp(struct lcfs_node_s *node)
...
+		return lcfs_load_node_from_file(dirfd, fname, buildflags);
+	}
+
+	if (fstatat(dirfd, fname, &sb, AT_SYMLINK_NOFOLLOW) < 0) {
+		return NULL;
+	}
+
+	/* Only regular files are tracked; directories can never be hardlinked
+	 * and are handled by recursing instead of via this function. */
+	if (!S_ISREG(sb.st_mode)) {
```
mike-sul: I think symlinks/FIFOs/device nodes are hardlinkable too, e.g. `ln -P`, and the composefs-rs implementation seems like processing all non-directory items  https://github.com/composefs/composefs-rs/blob/35a59a4500aeb5d1901586c782c8bbdb48f9a2a5/crates/composefs/src/fs.rs#L571.

mike-sul: I did a quick test and with a hardlinked symlink (ln -P) the digests still diverge.

mike-sul: It works now as expected. Thanks for a quick fix.

cgwalters: Thanks for the review!

